PR #12, opened 2019-04-01
package.json & tsconfig.json improvements

ColdenCullen:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "aws-crt",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "NodeJS bindings to the aws-c-* libraries",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -9,11 +9,21 @@
   ],
   "license": "Apache-2.0",
   "main": "./dist/lib/index.js",
+  "files": [
```
justinboswell: Why would the submodules be installed as dependencies?

ColdenCullen: Because if you don't get a pre-built binary, you need them to build. Getting them could probably be moved to the build step at some point, but it would be tricky.

justinboswell: Ok, I guess we'll go with it. It would be idea to avoid needing them eventually.

PR #13, opened 2019-07-11
Thread marshalling

JonathanHenson:
Mqtt now works on all supported platforms, reworked a few interfaces.


lib/mqtt.ts
```diff
@@ -66,18 +71,123 @@ export interface MqttSubscribeRequest extends MqttRequest {
...
+        this.params.connect_timeout = timeout;
+        return this;
+    }
+
+    build() {
+        if (this.params.client_id === undefined || this.params.host_name === undefined) {
+            throw 'client_id and endpoint are required';
+        }
+
+        if (this.tls_ctx_options === undefined) {
```
justinboswell: The other CRTs do not require TLS for all connections, mostly for testing.

scripts/deps_build.ts
```diff
@@ -43,8 +43,6 @@ const is_windows = process.platform == 'win32';
 const is_macos = process.platform == 'darwin';
 
 /* Capture the include path of Node dependencies */
```
justinboswell: Don't need this comment anymore

scripts/deps_build.ts
```diff
@@ -68,70 +66,29 @@ async function get_generator_string(): Promise<string | null> {
...
+            } 
 
             if (!vs_version_gen_str) {
                 console.error('CMake does not recognize an installed version of visual studio on your system.');
                 process.exit(1);
             }
 
             if (is_64bit) {
-                console.log('64bit version of python detected, using win64 builds')
+                console.log('64bit version of node detected, using win64 builds')
```
justinboswell: With VS2019, you can't add Win64, you have to instead add -A x64

source/mqtt_client.c
```diff
@@ -15,6 +15,8 @@
 #include "mqtt_client.h"
 #include "io.h"
 
+#include <assert.h>
```
justinboswell: assert.h is banned by clang-tidy. Why not use AWS_ASSERT?

source/module.c
```diff
@@ -19,6 +19,7 @@
 
 #include <aws/common/clock.h>
 
+#include <assert.h>
```
justinboswell: Why assert.h?

source/mqtt_client_connection.c
```diff
@@ -24,49 +24,400 @@
...
+
+    napi_value undefined = NULL;
+    napi_get_undefined(env, &undefined);
+    return undefined;
+}
+
+/************************************
+ * Connection object initialization
+ ************************************/
+static void s_dispatch_on_interupt(struct mqtt_nodejs_callback_context *context) {
```
justinboswell: interrupt

source/mqtt_client_connection.c
```diff
@@ -24,49 +24,400 @@
...
+    napi_handle_scope handle_scope = NULL;
+    napi_callback_scope cb_scope = NULL;
+
+    struct mqtt_nodejs_connection *node_connection = context->connection;
+    napi_env env = node_connection->env;
+
+    if (node_connection->on_connection_interupted) {
```
justinboswell: interrupted

source/mqtt_client_connection.c
```diff
@@ -24,49 +24,400 @@
...
+        napi_close_callback_scope(env, cb_scope);
+    }
+    if (handle_scope) {
+        napi_close_handle_scope(env, handle_scope);
+    }
+
+    aws_mem_release(context->allocator, context);
+}
+
+static void s_on_connection_interupted(struct aws_mqtt_client_connection *connection, int error_code, void *user_data) {
```
justinboswell: interrupted

source/mqtt_client_connection.c
```diff
@@ -106,34 +465,56 @@ napi_value mqtt_client_connection_new(napi_env env, napi_callback_info info) {
...
+            node_connection->connection,
+            s_on_connection_interupted,
+            node_connection,
+            s_on_connection_resumed,
+            node_connection);
     }
 
     napi_value node_external;
-    if (napi_create_external(env, node_connection, s_node_connection_finalize, NULL, &node_external)) {
+    if (napi_create_external(env, node_connection, NULL, NULL, &node_external)) {
```
justinboswell: No finalizer. So does this mean we're forcing people to ```close()``` their connections manually always?

source/mqtt_client_connection.c
```diff
@@ -555,85 +1003,60 @@ napi_value mqtt_client_connection_publish(napi_env env, napi_callback_info info)
...
+                s_raise_napi_error(env, s_callback_scope_open_failed);
                 goto cleanup;
             }
 
             napi_value recv;
             napi_value params[4];
-            if (napi_get_global(env, &recv) || napi_create_uint32(env, packet_id, &params[0]) ||
-                napi_create_string_utf8(env, (const char *)topic->ptr, topic->len, &params[1]) ||
-                napi_create_uint32(env, qos, &params[2]) || napi_create_int32(env, error_code, &params[3])) {
+            if (napi_get_global(env, &recv) || napi_create_int32(env, context->packet_id, &params[0]) ||
```
justinboswell: Just to be sure, the callback scope will clean up all these temporaries when we close it, is that correct?

test/test-load.ts
```diff
@@ -1,53 +1,58 @@
...
 
 async function main() {
     let bootstrap = new io.ClientBootstrap();
-    let tls_opt = io.TlsContextOptions.create_client_with_mtls("iot-certificate.pem.crt", "iot-private.pem.key")
-    tls_opt.override_default_trust_store(undefined, "AmazonRootCA1.pem")
-    tls_opt.alpn_list = "x-amzn-mqtt-ca"
-    let tls_ctx = new io.ClientTlsContext(tls_opt);
-    let client = new mqtt.Client(bootstrap, tls_ctx);
+    let config_builder = 
+    mqtt.AwsIotMqttConnectionConfigBuilder.new_mtls_builder_from_path("/home/ANT.AMAZON.COM/henso/source/4d169378e4.cert.pem",
```
justinboswell: Should make these arguments, no one else has your home directory :)

lib/mqtt.ts
JonathanHenson: this is the aws iot specific one. You can create a config object manually and use that if you want.

source/mqtt_client.c
JonathanHenson: Cause I didn't put the assert there. I'll go back and update.

source/mqtt_client_connection.c
JonathanHenson: yes, the finalizer is worthless.

test/test-load.ts
JonathanHenson: well, it's what it was before (something on colden's desktop). Yes, I should go back and do that.

source/mqtt_client_connection.c
JonathanHenson: This isn't particularly well documented in javascript, but the best I can tell is,

napi_create* creates a javascript object (identical to if you created it in javascript land) and will those be managed by javascript land when the user takes a reference to it.

https://nodejs.org/api/n-api.html#n_api_napi_create_object

Do you have any better thoughts?

ColdenCullen: If you have more questions, I highly suggest reading this: https://v8.dev/docs/embed

the tl;dr on this operation tho is that creating a callback scope creates a HandleScope (on the heap), meaning all new objects exist in that scope. When the scope is destroyed, all handles are cleaned up (note: this does NOT mean the memory is released, that's up to the GC).

source/mqtt_client_connection.c
```diff
@@ -24,49 +24,400 @@
...
+    napi_async_context callback_ctx;
+    struct aws_byte_buf payload_data;
+    struct mqtt_nodejs_connection *connection;
+    int error_code;
+    uint16_t packet_id;
+    struct aws_byte_buf topic;
+    bool session_present;
+    enum aws_mqtt_connect_return_code connect_code;
+    enum aws_mqtt_qos qos;
+    enum mqtt_nodejs_cb_type cb_type;
```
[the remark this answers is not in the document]
JonathanHenson: doesn't affect structure size, and you get mild readability improvements for dispatching, and a very minor perf improvement by having the IC warmed with the static table.

lib/mqtt.ts
```diff
@@ -66,18 +71,123 @@ export interface MqttSubscribeRequest extends MqttRequest {
 
 type Payload = string | Object | DataView;
 
-export class Connection {
+export class AwsIotMqttConnectionConfigBuilder {
```
ColdenCullen: I feel like this doesn't belong here, it belongs in the IoT SDK.

lib/mqtt.ts
```diff
@@ -17,6 +17,7 @@ import crt_native = require('./binding');
 
 import * as io from "./io";
 import { TextEncoder } from 'util';
+import ResourceSafety = require('./resource_safety')
```
ColdenCullen: You can do `import { ResourceSafe } from './resource_safety';` for more consistent style, then you don't have to qualify the access later in the file.

lib/resource_safety.ts
```diff
@@ -0,0 +1,26 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+ export interface ResourceSafe {
+     close(): void;
+ }
+
+ export async function using<T extends ResourceSafe>(resource : T, func: (resource: T) => void) {
```
ColdenCullen: Can you add a comment explaining what this is and how to use it?

package.json
```diff
@@ -42,6 +44,7 @@
     "typescript": "^3.2.2"
   },
   "dependencies": {
-    "node-pre-gyp": "^0.12.0"
+    "node-pre-gyp": "^0.12.0",
+    "yargs": "^13.2.4"
```
ColdenCullen: Since this is just used in a test script, I think it can safely be moved to `devDependencies`.

source/mqtt_client_connection.c
```diff
@@ -24,49 +24,403 @@
...
+        napi_delete_reference(env, node_connection->on_connection_resumed);
+    }
+
+    /* since closing this handle is asynchronous, we need to wait for the callback to clean
+       up the connection. */
+    uv_close((uv_handle_t *)&node_connection->async_handle, s_async_handle_close_cb);
+
+    napi_value undefined = NULL;
+    napi_get_undefined(env, &undefined);
+    return undefined;
```
ColdenCullen: You can also just return `NULL` here and it'll translate to undefined.

lib/mqtt.ts
JonathanHenson: import { using } from .....

lib/mqtt.ts
JonathanHenson: eeeh, let's revisit this later. I did move it to a different file though.

lib/resource_safety.ts
```diff
@@ -13,14 +13,26 @@
...
-     } finally {
-         resource.close();
-     }
- }
\ No newline at end of file
+/*
+ * Use this function to create a resource in an async context. This will make sure the 
+ * resources are cleaned up before returning.
+ * 
+ * example:   await using(res = new SomeResource()) {
```
justinboswell: this isn't the right syntax, it should take the lambda as the second argument in ```using()```

PR #17, opened 2019-08-20
Wrapped crypto-js with the CRT hash interface, added unit testing

justinboswell:


lib/browser/crypto.ts
```diff
@@ -13,4 +13,48 @@
...
+        const encoder = new TextEncoder();
+        const bytes = encoder.encode(truncated);
+        return new DataView(bytes);
+    }
+}
+
+export function hash_md5(data: Hashable, truncate_to?: number): DataView {
+    const md5 = new Md5Hash();
+    md5.update(data);
+    return md5.digest();
```
bretambrose: Does truncate_to get magically used here?

lib/browser/crypto.ts
```diff
@@ -13,4 +13,48 @@
...
+import { Hashable } from "../common/crypto";
+
+export class Md5Hash {
+    private hash?: Crypto.WordArray;
+    
+    update(data: Hashable) {
+        this.hash = Crypto.MD5(data.toString(), this.hash ? this.hash.toString() : undefined);
+    }
+
+    digest(truncate_to?: number): DataView {
```
bretambrose: I have to confess I like "finalize" better than "digest" both for consistency as well as action vs. noun.

lib/browser/crypto.ts
```diff
@@ -13,4 +13,48 @@
...
+    }
+}
+
+export function hash_md5(data: Hashable, truncate_to?: number): DataView {
+    const md5 = new Md5Hash();
+    md5.update(data);
+    return md5.digest();
+}
+
+export class Sha256Hmac {
```
bretambrose: Does this need a top-level wrapper function?

lib/browser/crypto.ts
justinboswell: No, it gets magically forgotten by the implementer, thanks!

lib/browser/crypto.ts
justinboswell: Now is the time, I can fix it!

lib/browser/crypto.ts
justinboswell: yes it does, I deleted it by accident, aaaahhhhh

tsconfig.browser.json
```diff
@@ -5,6 +5,7 @@
     },
     "include": [
         "lib/*.ts",
```
ColdenCullen: This should be empty now, right?

tsconfig.json
```diff
@@ -2,6 +2,7 @@
     "extends": "./tsconfig.base",
     "include": [
         "lib/*.ts",
```
ColdenCullen: Same

tsconfig.browser.json
justinboswell: Fixed.

tsconfig.json
justinboswell: Fixed

PR #21, opened 2019-08-29
HTTP Bindings

justinboswell:

### DIFF
--- a/lib/native/http.ts
+++ b/lib/native/http.ts
@@ -71,7 +71,7 @@ export class HttpClientConnection extends HttpConnection {
                 host_name,
                 port,
                 socket_options.native_handle(),
-                tls_ctx
+                tls_ctx ? tls_ctx.native_handle() : undefined
             )
         });
     }
--- a/lib/native/http.ts
+++ b/lib/native/http.ts
@@ -18,31 +18,69 @@ import { NativeResource } from "./native_resource";
 import { ResourceSafe } from '../common/resource_safety';
 import { ClientBootstrap, ClientTlsContext, SocketOptions } from './io';
 
-type ConnectionCallback = (error_code: Number) => void;
-
 export class HttpConnection extends NativeResource implements ResourceSafe {
 
-    constructor(
-        protected bootstrap: ClientBootstrap,
-        protected on_connection_setup: ConnectionCallback | undefined,
-        protected on_connection_shutdown: ConnectionCallback | undefined,
-        host_name: String,
-        port: Number,
-        protected socket_options: SocketOptions,
-        protected tls_ctx: ClientTlsContext)
+    protected constructor(native_handle: any)
     {
-        super(crt_native.http_connection_new(
-            bootstrap.native_handle(),
-            on_connection_setup,
-            on_connection_shutdown,
-            host_name,
-            port,
-            socket_options.native_handle(),
-            tls_ctx.native_handle()
-        ));
+        super(native_handle);
     }
 
     close() {
         crt_native.http_connection_close(this.native_handle());
+    }
+}
+
+type ClientConnectionCallback = (connection: HttpClientConnection, error_code: Number) => void;
+
+export class HttpClientConnection extends HttpConnection {
+    static create(
+        bootstrap: ClientBootstrap,
+        on_connection_setup: ClientConnectionCallback | undefined,
+        on_connection_shutdown: ClientConnectionCallback | undefined,
+        host_name: String,
+        port: Number,
+        socket_options: SocketOptions,
+        tls_ctx: ClientTlsContext) {
+        
+        return new Promise<HttpConnection>((resolve) => {
+            let connection: HttpClientConnection;
+            const on_setup = (native_connection: any, error_code: Number) => {
+                connection = new HttpClientConnection(
+                    native_connection,
+                    bootstrap,
+                    socket_options,
+                    tls_ctx);
+                if (on_connection_setup) {
+                    on_connection_setup(connection, error_code);
+                }
+                resolve(connection);
+            }
+
+            const on_shutdown = (_native_connection: any, error_code: Number) => {
+                if (on_connection_shutdown) {
+                    on_connection_shutdown(connection, error_code);
+                }
+            }
+
+            // create new connection, connection will be delivered via on_setup
+            // errors during construction will be thrown
+            crt_native.http_connection_new(
+                bootstrap.native_handle(),
+                on_setup,
+                on_shutdown,
+                host_name,
+                port,
+                socket_options.native_handle(),
+                tls_ctx.native_handle()
+            )
+        });
+    }
+
+    protected constructor(
+        native_handle: any,
+        protected bootstrap: ClientBootstrap,
+        protected socket_options: SocketOptions,
+        protected tls_ctx: ClientTlsContext) {
+        super(native_handle);
     }
 }
--- a/lib/native/http.ts
+++ b/lib/native/http.ts
@@ -40,9 +40,9 @@ export class HttpClientConnection extends HttpConnection {
         host_name: String,
         port: Number,
         socket_options: SocketOptions,
-        tls_ctx: ClientTlsContext) {
+        tls_ctx?: ClientTlsContext) : Promise<HttpClientConnection> {
         
-        return new Promise<HttpConnection>((resolve) => {
+        return new Promise<HttpClientConnection>((resolve) => {
             let connection: HttpClientConnection;
             const on_setup = (native_connection: any, error_code: Number) => {
                 connection = new HttpClientConnection(
@@ -71,7 +71,7 @@ export class HttpClientConnection extends HttpConnection {
                 host_name,
                 port,
                 socket_options.native_handle(),
-                tls_ctx.native_handle()
+                tls_ctx
             )
         });
     }
@@ -80,7 +80,7 @@ export class HttpClientConnection extends HttpConnection {
         native_handle: any,
         protected bootstrap: ClientBootstrap,
         protected socket_options: SocketOptions,
-        protected tls_ctx: ClientTlsContext) {
+        protected tls_ctx?: ClientTlsContext) {
         super(native_handle);
     }
 }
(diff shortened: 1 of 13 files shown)


scripts/deps_build.ts
```diff
@@ -147,11 +147,11 @@ async function build_dependency(lib_name: string, ...cmake_args: string[]) {
         '-DBUILD_SHARED_LIBS=OFF',
         '-DBUILD_TESTING=OFF',
         '-DCMAKE_INSTALL_LIBDIR=' + lib_dir,
-        '-DCMAKE_BUILD_TYPE=Release',
+        '-DCMAKE_BUILD_TYPE=Debug',
         cmake_args.join(' '),
         lib_source_dir,
     ].join(' ');
-    const build_cmd = ['cmake', '--build', './', '--config', 'release', '--target', 'install'].join(' ');
+    const build_cmd = ['cmake', '--build', './', '--config', 'debug', '--target', 'install'].join(' ');
```
graebm: Probably want this toggleable base on some build param?
Or did you just not mean to check this in

lib/native/http.ts
```diff
@@ -18,31 +18,69 @@ import { NativeResource } from "./native_resource";
...
+    {
+        super(native_handle);
+    }
+
+    close() {
+        crt_native.http_connection_close(this.native_handle());
+    }
+}
+
+type ClientConnectionCallback = (connection: HttpClientConnection, error_code: Number) => void;
```
graebm: Just a note: The bindings need something like an ErrorCode class so they can make any sense of these ints. I started writing one yesterday but put it on the backburner when I realized I should focus on HTTP first.

lib/native/http.ts
```diff
@@ -18,31 +18,69 @@ import { NativeResource } from "./native_resource";
...
-            port,
-            socket_options.native_handle(),
-            tls_ctx.native_handle()
-        ));
+        tls_ctx: ClientTlsContext) {
+        
+        return new Promise<HttpConnection>((resolve) => {
+            let connection: HttpClientConnection;
+            const on_setup = (native_connection: any, error_code: Number) => {
+                connection = new HttpClientConnection(
```
graebm: whoa, you can capture a variable that hasn't been set yet.

black

magic

source/http_connection.c
```diff
@@ -14,33 +14,37 @@
...
 #include "uv_interop.h"
 
 #include <aws/http/connection.h>
 #include <aws/io/tls_channel_handler.h>
 
-struct http_nodejs_connection {
+struct http_connection_binding {
     struct aws_http_connection *connection;
+    napi_value node_external;
     struct aws_tls_connection_options tls_options;
```
graebm: there's no need to keep the `aws_tls_connection_options` struct as a permanent member. It's all copied when aws_http_client_connect() is called.

lib/native/http.ts
```diff
@@ -18,31 +18,69 @@ import { NativeResource } from "./native_resource";
...
 
-    constructor(
+    protected constructor(native_handle: any)
+    {
+        super(native_handle);
+    }
+
+    close() {
+        crt_native.http_connection_close(this.native_handle());
+    }
```
graebm: You're never calling `aws_http_connection_release()` anywhere.

`aws_http_connection_close()` tells the connection to shutdown the channel, if it's not already shut down, but this won't necessarily clean up the `aws_http_connection` struct.

The connection isn't cleaned up until BOTH `aws_http_connection_release()` is called AND the on_shutdown callback has fired.

If this `close()` function is supposed to mean "finalize resource" then you probably want it to call `aws_http_connection_release()` instead. But you'll still need to keep the JS object alive until on_shutdown has fired (unless it's fired already). 

```
/**
 * Users must release the connection when they are done with it.
 * The connection's memory cannot be reclaimed until this is done.
 * If the connection was not already shutting down, it will be shut down.
 *
 * Users should always wait for the on_shutdown() callback to be called before releasing any data passed to the
 * http_connection (Eg aws_tls_connection_options, aws_socket_options) otherwise there will be race conditions between
 * http_connection shutdown tasks and memory release tasks, causing Segfaults.
 */
AWS_HTTP_API
void aws_http_connection_release(struct aws_http_connection *connection);
```

scripts/deps_build.ts
justinboswell: I'll make it a param or something. This is temp.

lib/native/http.ts
justinboswell: That one took me a minute to come up with. Futures create a time paradox.

source/http_connection.c
justinboswell: oh fabulous, will fix.

lib/native/http.ts
justinboswell: Yeah, I haven't been able to test this yet, and didn't get there yet. This was mostly so you could see the ts code. I'll get it sorted though.

lib/native/http.ts
justinboswell: I added an Error class that converts int -> string

lib/native/http.ts
```diff
@@ -18,31 +18,69 @@ import { NativeResource } from "./native_resource";
 import { ResourceSafe } from '../common/resource_safety';
 import { ClientBootstrap, ClientTlsContext, SocketOptions } from './io';
 
-type ConnectionCallback = (error_code: Number) => void;
-
 export class HttpConnection extends NativeResource implements ResourceSafe {
 
-    constructor(
+    protected constructor(native_handle: any)
+    {
```
JonathanHenson: style consistency

source/http_connection.c
```diff
@@ -222,23 +278,18 @@ napi_value aws_napi_http_connection_close(napi_env env, napi_callback_info info)
...
+    if (napi_get_value_external(env, node_args[0], (void **)&binding)) {
+        napi_throw_error(env, NULL, "Failed to extract http_connection_binding from external");
         return NULL;
     }
 
-    if (node_connection->connection) {
-        aws_http_connection_close(node_connection->connection);
+    if (binding->connection) {
+        aws_http_connection_close(binding->connection);
+        aws_http_connection_release(binding->connection);
```
graebm: trivial: don't need to call `aws_http_connection_close()` release() does that for you

lib/native/http.ts
```diff
@@ -18,31 +18,69 @@ import { NativeResource } from "./native_resource";
...
-            socket_options.native_handle(),
-            tls_ctx.native_handle()
-        ));
+        tls_ctx?: ClientTlsContext) : Promise<HttpClientConnection> {
+        
+        return new Promise<HttpClientConnection>((resolve) => {
+            let connection: HttpClientConnection;
+            const on_setup = (native_connection: any, error_code: Number) => {
+                connection = new HttpClientConnection(
+                    native_connection,
```
graebm: If `error_code` is non-zero then `native_connection` is NULL and the Promise should resolve with an exception.

lib/native/http.ts
justinboswell: This is fixed in my follow up work.

source/http_connection.c
```diff
@@ -108,22 +144,25 @@ napi_value aws_napi_http_connection_new(napi_env env, napi_callback_info info) {
...
-            return NULL;
-        }
+    if (aws_napi_is_null_or_undefined(env, node_args[1])) {
+        napi_throw_error(env, NULL, "2nd argument (on_connection_setup) must be a callback");
+        return NULL;
+    }
+    if (aws_napi_callback_init(
+            &on_connection_setup,
+            env,
+            node_args[1],
```
graebm: We don't clean up the result of THIS `aws_napi_callback_init()` call if the NEXT call fails a few lines down

source/http_connection.c
```diff
@@ -158,59 +197,76 @@ napi_value aws_napi_http_connection_new(napi_env env, napi_callback_info info) {
...
     options.on_shutdown = s_http_on_connection_shutdown;
-    options.user_data = node_connection;
-    aws_tls_connection_options_init_from_ctx(&node_connection->tls_options, tls_ctx);
-    node_connection->tls_options.server_name = host_name;
-    options.tls_options = &node_connection->tls_options;
+    options.user_data = binding;
+    struct aws_tls_connection_options tls_options;
+    AWS_ZERO_STRUCT(tls_options);
+    if (tls_ctx) {
+        aws_tls_connection_options_init_from_ctx(&tls_options, tls_ctx);
```
graebm: tls_options should be cleaned up with `aws_tls_connection_options_clean_up()` before this function returns, whether or not the final aws_http_client_connect() call succeeds.

It's ok to call clean_up on the zeroed-out options, even if you never init'ed them

source/http_connection.c
```diff
@@ -158,59 +197,76 @@ napi_value aws_napi_http_connection_new(napi_env env, napi_callback_info info) {
...
-    node_connection->on_setup = on_connection_setup;
-    node_connection->on_shutdown = on_connection_shutdown;
+    if (napi_create_reference(env, node_external, 1, &binding->node_external)) {
+        napi_throw_error(env, NULL, "Failed to reference node_external");
+        goto create_external_failed;
+    }
 
+    binding->env = env;
+    binding->uv_context = aws_uv_context_get_default();
+    aws_uv_context_acquire(binding->uv_context, env);
```
graebm: if `aws_uv_context_acquire()` can't fail, let's make it a `void` function

source/http_connection.c
justinboswell: I've fixed this in a follow-up

source/http_connection.c
justinboswell: It doesn't need to be cleaned up. A failed init results in a clean structure.

lib/native/http.ts
justinboswell: Good point, fixed.

source/http_connection.c
justinboswell: So, it can fail, let me see what I can do about that.

source/http_connection.c
justinboswell: Fixed

PR #25, opened 2019-09-12
Input Stream implementation, body send changes

justinboswell:

### DIFF
--- a/source/io.c
+++ b/source/io.c
@@ -516,7 +516,7 @@ static int s_input_stream_get_status(struct aws_input_stream *stream, struct aws
 static int s_input_stream_get_length(struct aws_input_stream *stream, int64_t *out_length) {
     (void)stream;
     (void)out_length;
-    return AWS_ERROR_UNIMPLEMENTED;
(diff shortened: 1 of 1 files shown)


source/io.c
```diff
@@ -421,3 +423,214 @@ napi_value aws_napi_io_socket_options_new(napi_env env, napi_callback_info info)
...
+    (void)out_length;
+    return aws_raise_error(AWS_ERROR_UNIMPLEMENTED);
+}
+
+static void s_input_stream_clean_up(struct aws_input_stream *stream) {
+    struct aws_napi_input_stream_impl *impl = stream->impl;
+    struct aws_allocator *allocator = impl->buffer.allocator;
+    aws_mutex_clean_up(&impl->mutex);
+    aws_byte_buf_clean_up(&impl->buffer);
+    aws_mem_release(allocator, impl);
```
graebm: The `aws_input_stream_clean_up()` "virtual" function calls `aws_mem_release(stream->allocator, stream);` on the stream.

IT SHOULDN'T. WE SHOULD FIX THIS NOW.

This was leading to a double-free in the Python code, and this code now has the same issue

PR #28, opened 2019-09-16
Docs

justinboswell:
Note that MQTT needs a docs pass, and I need to add docs once HttpConnection is converted to using events.

### DIFF
--- a/lib/common/http.ts
+++ b/lib/common/http.ts
@@ -17,7 +17,8 @@ import { InputStreamBase } from "./io";
 
 type HttpHeader = [string, string];
 
-/** Encapsulates an HTTP header block. Stores all headers in their original case format,
+/** 
+ * Encapsulates an HTTP header block. Stores all headers in their original case format,
  * but allows for case-insensitive header lookup.
  */
 export class HttpHeaders {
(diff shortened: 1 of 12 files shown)


lib/common/http.ts
```diff
@@ -17,16 +17,24 @@ import { InputStreamBase } from "./io";
...
 
+    /** Construct from a collection of [name, value] pairs */
     constructor(headers: HttpHeader[] = []) {
         for (const header of headers) {
             this.add(header[0], header[1]);
         }
     }
 
+    /** Add a name/value pair 
+     * @param name - The header name
```
graebm: trivial: I mildly prefer:
```
/**
 * First line
 * next line
 */
```
over
```
/** First line
 * next line
 */
```

but I'm also whatever

lib/native/io.ts
```diff
@@ -15,32 +15,60 @@
 
 import crt_native = require('./binding');
 import { NativeResource } from "./native_resource";
-import { ResourceSafe } from '../common/resource_safety';
-import { InputStreamBase } from '../common/io';
+import { InputStreamBase, TlsVersion } from '../common/io';
 import { Readable } from 'stream';
+export { TlsVersion } from '../common/io';
```
ColdenCullen: double import?

ColdenCullen: JK fuck exports

justinboswell: Nope, import and export are separate/different.

PR #30, opened 2019-09-18
Updated MQTT to match HTTP's api and patterns

justinboswell:


lib/browser/mqtt.ts
```diff
@@ -12,18 +12,20 @@
...
 export { QoS, Payload, MqttRequest, MqttSubscribeRequest } from "../common/mqtt";
 
 export type WebsocketOptions = WebsocketUtils.WebsocketOptions;
 export type AWSCredentials = WebsocketUtils.AWSCredentials;
 
-export interface ConnectionConfig {
+export interface MqttConnectionConfig {
     client_id: string;
     host_name: string;
     connect_timeout: number;
```
graebm: trivial: why isn't connect_timeout optional like most other options here?

lib/native/mqtt.ts
```diff
@@ -54,30 +48,67 @@ export interface ConnectionConfig {
...
+    /**
+     * Emitted when the connection is dropped unexpectedly. The error will contain the error
+     * code and message.
+     */
+    on(event: 'interrupt', listener: (error: CrtError) => void): this;
+
+    /**
+     * Emitted when the connection reconnects. Only triggers on connections after the initial one.
+     */
+    on(event: 'resume', listener: (return_code: number, session_present: boolean) => void): this;
```
graebm: why does 'resume' pass `return_code` but 'connect' doesn't?

samples/browser/pub_sub/index.ts
```diff
@@ -51,50 +51,48 @@ async function connect_websocket(credentials: AWS.CognitoIdentityCredentials) {
...
-            reject(reason);
         });
+        connection.on('interrupt', (error: CrtError) => {
+            log(`Connection interrupted: error=${error}`);
+        });
+        connection.on('resume', (return_code: number, session_present: boolean) => {
+            log(`Resumed: rc: ${return_code} existing session: ${session_present}`)
+        });
+        connection.on('disconnect', () => {
+            log('Disconnected');
```
graebm: THIS IS ALL SO SIMPLE AND BEAUTIFUL

lib/native/error.ts
```diff
@@ -21,8 +21,12 @@ import { isNumber } from 'util';
  * a human-readable string.
  */
 export class CrtError extends Error {
+    public readonly error_code?: number;
     constructor(err: any) {
```
graebm: probably also want to save the AWS_ERROR_ENUM_NAME because honestly if they're handling specific errors that's the easiest one to check against. ex:
`if (error.name == "AWS_ERROR_SOCKET_TIMEOUT")`

But ugh.. an enum would be better.
And a definitive list of possible values would be EVEN BETTER but that would involve some codegen... or an O(N) crawl over possible error values at boot to build up some dynamic list.

ramble ramble

lib/native/io.ts
```diff
@@ -145,19 +145,23 @@ export class TlsContextOptions {
...
     public min_tls_version: TlsVersion = TlsVersion.Default;
     /** Path to a single file with all trust anchors in it, in PEM format */
-    public ca_file?: string = undefined;
+    public ca_file?: string;
     /** Path to directory containing trust anchors. Only used on Unix-style systems. */
-    public ca_path?: string = undefined;
+    public ca_path?: string;
     /** Semi-colon delimited list of ALPN protocols to be used on platforms which support ALPN */
-    public alpn_list?: string = undefined;
+    public alpn_list?: string;
```
graebm: don't do it in this PR
but maybe you want to copy what I did today in python:
- rename `ca_path` to `ca_dirpath`
- rename all the other `_path` or `_file` args/vars to `_filepath`
- make `alpn_list` an actual list instead of semicolon delimited string

samples/browser/pub_sub/index.ts
justinboswell: :1st_place_medal:

lib/native/io.ts
justinboswell: That sounds like a magnificent follow up

lib/native/error.ts
justinboswell: I can add that. Can I do it as a follow up before this PR kills someone?

lib/native/mqtt.ts
justinboswell: If you connected, the only possible return code is 0 AFAIK. So we can either leave it out, or just pass 0s to both.

lib/browser/mqtt.ts
justinboswell: No reason, totally fixable.

lib/native/mqtt.ts
graebm: you know what. python has the same thing.
I don't know what's up with return_code either
not sure exactly what state you're in if that's non-zero

lib/native/error.ts
graebm: yes. absolutely

test/mqtt.ts
```diff
@@ -0,0 +1,138 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import * as AWS from 'aws-sdk';
```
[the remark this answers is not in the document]
justinboswell: yes, it's in the ```devDependencies``` in package.json

PR #36, opened 2019-09-27
Fixes made while testing the IoT SDK

justinboswell:


source/module.c
```diff
@@ -284,8 +345,9 @@ int aws_napi_callback_dispatch(struct aws_napi_callback *cb, void *user_data) {
     }
     AWS_FATAL_ASSERT(num_params < AWS_ARRAY_SIZE(params));
 
-    if (napi_make_callback(env, cb->async_context, this_object, node_function, num_params, params, NULL)) {
-        napi_throw_error(env, NULL, "Callback invocation failed");
+    napi_status call_result = napi_make_callback(env, cb->async_context, this_object, node_function, num_params, params, NULL);
```
justinboswell: Some context:
If a callback throws an uncaught exception, the call result will be ```napi_pending_exception``` and no callbacks will be delivered thereafter. I want to research how to clean that up/remedy the situation as a follow-up. Therefore, I added default error handlers to our connection/stream classes to prevent a throw (which is what happens if there is no error handler). I also swapped this to log, since the throw it was doing before would never be seen inside of node.

lib/native/aws_iot.ts
```diff
@@ -70,34 +71,69 @@ export class AwsIotMqttConnectionConfigBuilder {
...
+    }
 
+    /**
+     * Overrides the default system trust store.
+     * @param ca - Buffer containing all trust CAs, in PEM format
+     */
+    with_certificate_authority(ca: string) {
+        if (this.tls_ctx_options) {
+            this.tls_ctx_options.override_default_trust_store(ca);
+        }
```
graebm: It's a little weird that this would silently have no effect

lib/native/aws_iot.ts
```diff
@@ -70,34 +71,69 @@ export class AwsIotMqttConnectionConfigBuilder {
...
+     * Overrides the default system trust store.
+     * @param ca_dirpath - Only used on Unix-style systems where all trust anchors are
+     * stored in a directory (e.g. /etc/ssl/certs).
+     * @param ca_filepath - Single file containing all trust CAs, in PEM format
+     */
+    with_certificate_authority_from_path(ca_dirpath?: string, ca_filepath?: string) {
         if (this.tls_ctx_options !== undefined) {
-            this.tls_ctx_options.override_default_trust_store(ca_path, ca_file);
+            this.tls_ctx_options.override_default_trust_store_from_path(ca_dirpath, ca_filepath);
         }
```
graebm: likewise, weird that this silently fails to have any effect

lib/native/http.ts
```diff
@@ -28,6 +28,7 @@ export class HttpConnection extends NativeResourceMixin(BufferedEventEmitter) im
     protected constructor(native_handle: any) {
         super();
         this._super(native_handle);
+        this.on('error', () => { }); // suppress error throwing
```
graebm: Node's typical [behavior](https://nodejs.org/api/events.html#events_error_events) is:
> If an EventEmitter does not have at least one listener registered for the 'error' event, and an 'error' event is emitted, the error is thrown, a stack trace is printed, and the Node.js process exits.

Users expect it to throw

justinboswell: I've fixed this with new error handling, now that I understand the problem.

source/module.c
justinboswell: This was resolved by new error handling

source/module.c
```diff
@@ -237,6 +298,105 @@ int aws_napi_callback_clean_up(struct aws_napi_callback *cb) {
     return AWS_OP_SUCCESS;
 }
 
+static void s_handle_failed_callback(napi_env env, napi_status status) {
+    /* Figure out if there's an exception pending, if so, no callbacks will ever succeed again until it's cleared */
+    bool pending_exception = status == napi_pending_exception;
+    if ((status = napi_is_exception_pending(env, &pending_exception))) {
```
graebm: it looks like you're setting an initial value for `pending_exception`, then passing it as an out-param?
is that on purpose?

source/module.c
```diff
@@ -237,6 +298,105 @@ int aws_napi_callback_clean_up(struct aws_napi_callback *cb) {
     return AWS_OP_SUCCESS;
 }
 
+static void s_handle_failed_callback(napi_env env, napi_status status) {
```
graebm: `status` gets passed in, but it's immediately overwritten by `if ((status = napi_xyz()))` calls
is that on purpose?

source/module.c
justinboswell: So, if the status is napi_pending_exception, there is DEFINITELY an exception. However, the other statuses could ALSO mean there is an exception, which is why you have to ask. It doesn't really matter, unless ```napi_is_exception_pending()``` fails somehow. It's just the default value, I could have easily used ```false``` or ```true```.

source/module.c
justinboswell: Yes. The status parameter is literally only there to be the initial value of pending_exception, which is in turn only set to that in case the next call fails and you're debugging. Otherwise, it's all meaningless.

PR #38, opened 2019-09-30
Logging and lifetime/error reporting fixes

justinboswell:


lib/native/http.ts
```diff
@@ -291,6 +291,9 @@ export class HttpClientConnectionManager extends NativeResource {
                         handle
                     );
                     this.connections.set(handle, connection as HttpClientConnection);
+                    connection.on('close', () => {
```
justinboswell: Bug 1: connections lived forever, even when they were dead, causing them to leak

source/http_connection.c
```diff
@@ -31,13 +31,13 @@ struct http_connection_binding {
...
 /* finalizer called when node cleans up this object */
-static void s_http_connection_binding_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
+static void s_http_connection_from_manager_binding_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
     (void)env;
     (void)finalize_hint;
     struct http_connection_binding *binding = finalize_data;
     struct aws_allocator *allocator = aws_default_allocator();
 
-    aws_http_connection_release(binding->connection);
+    /* no release call, the http_client_connection_manager has already released it */
```
justinboswell: Bug 2: Double release of connections released by the connection manager

source/module.h
```diff
@@ -59,4 +60,45 @@ int aws_napi_callback_init(
...
+    struct aws_napi_logger_ctx *logger;
+};
+
+#define _AWS_NAPI_ERROR_MSG(call, file, line) "N-API call failed: " #call " @ " file "(" #line ")"
+#define _AWS_NAPI_PASTE(x) x
+
+/*
+ * AWS_NAPI_CALL(env, napi_xxx(args...), { return NULL; }) will ensure that a failed result is logged as an error
+ */
+#define AWS_NAPI_CALL(env, call, on_fail)                                                                              \
```
justinboswell: I'm happy to do a follow up where all napi calls are wrapped in these 2 macros, but this PR is already a nightmare.

source/mqtt_client_connection.c
```diff
@@ -99,6 +98,11 @@ napi_value aws_napi_mqtt_client_connection_close(napi_env env, napi_callback_inf
     aws_napi_callback_clean_up(&binding->on_connection_resumed);
     napi_delete_reference(env, binding->node_external);
 
+    /* destroy the native connection, which will destroy the subscriptions and
+     * queue destruction of all bound callbacks
+     */
+    aws_mqtt_client_connection_destroy(binding->connection);
```
justinboswell: Bug 4: We were not destroying mqtt connections early enough in the cycle, so they could be holding callbacks in the sub trie, and that would prevent node from shutting down.

source/uv_interop.c
```diff
@@ -106,17 +106,23 @@ static void s_uv_closed(uv_handle_t *handle) {
     aws_mem_release(ctx->allocator, ctx);
 }
 
-void aws_uv_context_release(struct aws_uv_context *ctx) {
-    aws_mutex_lock(&ctx->command_queue.mutex);
-    /* For now, don't bother supporting a final flush, it shouldn't be necessary, as when refs are
-       dropped the owning object should be on its way to death */
+static void s_context_release(void *user_data) {
+    struct aws_uv_context *ctx = user_data;
+    /* this should be the last command in the queue, so the queue should be empty */
```
justinboswell: Remember when I thought we wouldn't need this? Oops.

For now, we do. If we like this mechanism, ```napi_threadsafe_function``` is identical to this in functionality, so I can just transition away from this code and use the node-provided API. We didn't originally because we couldn't depend on NAPI > 1, but we have to now, as those versions of node are past LTS.

source/logger.c
```diff
@@ -0,0 +1,296 @@
...
+    struct log_message *msg = aws_mem_calloc(&ctx->buffer_allocator, 1, sizeof(struct log_message));
+    msg->message = message;
+
+    /* queue up the message to be logged next time the async work runs */
+    aws_mutex_lock(&ctx->msg_queue.mutex);
+    aws_linked_list_push_back(&ctx->msg_queue.messages, &msg->node);
+    aws_mutex_unlock(&ctx->msg_queue.mutex);
+
+    /* pin the threadsafe function until we try to call it */
+    AWS_NAPI_ENSURE(ctx->env, napi_acquire_threadsafe_function(ctx->log_drain));
```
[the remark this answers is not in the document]
justinboswell: It is, when it is called, in ```s_threadsafe_log_call()```

source/logger.c
```diff
@@ -0,0 +1,296 @@
...
+    };
+    struct aws_ring_buffer *buffer = allocator->impl;
+    aws_ring_buffer_release(buffer, &buf);
+}
+
+static void *s_ring_buffer_mem_calloc(struct aws_allocator *allocator, size_t num, size_t size) {
+    void *mem = s_ring_buffer_mem_acquire(allocator, num * size);
+    memset(mem, 0, size);
+    return mem;
+}
```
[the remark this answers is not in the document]
justinboswell: yeah, I can do that, mind if I do it as a follow up?

PR #39, opened 2019-10-17
Implement MQTT client in terms of socket_options

ColdenCullen:

### DIFF
--- a/lib/browser/io.ts
+++ b/lib/browser/io.ts
@@ -34,9 +34,9 @@ export class InputStream implements InputStreamBase {
 */
 export class SocketOptions {
     constructor(
-        public type: SocketType = SocketType.STREAM,
-        public domain: SocketDomain = SocketDomain.IPV6,
-        public connect_timeout_ms: number = 5000,
+        public type = SocketType.STREAM,
+        public domain = SocketDomain.IPV6,
+        public connect_timeout_ms = 5000,
         public keepalive = false,
         public keep_alive_interval_sec = 0,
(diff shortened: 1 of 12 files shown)


lib/browser/io.ts
```diff
@@ -23,6 +23,23 @@ type BodyData = string | object | ArrayBuffer | ArrayBufferView | Blob | File;
...
+
+/**
+ * Standard Berkeley socket style options.
+ *
+ * Provided for compatibility with nodejs, but this version is largely unused.
+*/
+export class SocketOptions {
+    constructor(
+        public type: SocketType = SocketType.STREAM,
+        public domain: SocketDomain = SocketDomain.IPV6,
```
justinboswell: Shouldn't this be IPV4?

lib/browser/io.ts
```diff
@@ -23,6 +23,23 @@ type BodyData = string | object | ArrayBuffer | ArrayBufferView | Blob | File;
...
+}
+
+/**
+ * Standard Berkeley socket style options.
+ *
+ * Provided for compatibility with nodejs, but this version is largely unused.
+*/
+export class SocketOptions {
+    constructor(
+        public type: SocketType = SocketType.STREAM,
```
justinboswell: Don't need type on these, since you're initializing them.

lib/browser/io.ts
ColdenCullen: It's IPV6 in Python, so I matched that. I could go either way on this though.

PR #40, opened 2019-10-17
Converted callbacks to use napi_threadsafe_function

justinboswell:


source/logger.c
```diff
@@ -131,7 +134,15 @@ static void s_ring_buffer_mem_release(struct aws_allocator *allocator, void *ptr
         .len = 0,
     };
     struct aws_ring_buffer *buffer = allocator->impl;
-    aws_ring_buffer_release(buffer, &buf);
+
+    /* see if the memory comes from the ring buffer */
+    if (aws_ring_buffer_buf_belongs_to_pool(buffer, &buf)) {
+        aws_ring_buffer_release(buffer, &buf);
+        return;
```
ColdenCullen: random style thing, instead of returning here, just put the `aws_meme_release` call in an `else` block.

source/http_stream.c
```diff
@@ -16,65 +16,65 @@
...
+        for (size_t idx = 0; idx < num_headers; ++idx) {
+            struct aws_http_header header;
+            aws_http_message_get_header(response, &header, idx);
+
+            napi_value node_header = NULL;
+            AWS_NAPI_ENSURE(env, napi_create_array(env, &node_header));
+
+            napi_value node_name = NULL;
+            napi_value node_value = NULL;
+            AWS_NAPI_ENSURE(
```
graebm: AWS_NAPI_ENSURE is a bold approach to error handling

I've been wondering how this kind of thing should be handled (errors while prepping a callback) in the python CRT.
I've definitely been bit where elasticurl was failing in CI because we weren't waiting for the connection to finish shutdown (which should be fine in Python), so the on_shutdown in C tried to call on_shutdown in Python, which failed, so I AWS_FATAL_ASSERTed, which made it look like elasticurl was failing.

Is Node culture that an application should 100% close things down before exiting culture?

justinboswell: This is only for use in the case where throwing an exception or something is not a viable alternative for handling a failure. In this case, we're in the node thread, about to call a node function, and we're pushing the arguments onto the local scope. We cannot throw an exception, because it is physically impossible for someone to catch it, so it would just prevent execution until our exception clearing code runs (see module.c). Therefore, the only option is to crash hard, there's no real recovery. Also, if node fails to create an int or a string, it's not going to succeed at much after that point.

If you look at the node-addon-examples github repo, this pattern is basically what they do. They assert on everything that is very unlikely to fail, and catch on things that are prone to failure. I have tried where possible to follow that example. That's why we have ```AWS_NAPI_ENSURE``` and ```AWS_NAPI_CALL```.

PR #43, opened 2019-10-29
Websockets & Proxies

ColdenCullen:

### DIFF
--- a/lib/common/http.ts
+++ b/lib/common/http.ts
@@ -17,7 +17,7 @@ import { InputStreamBase } from "./io";
 
 type HttpHeader = [string, string];
 
-/** 
+/**
  * Encapsulates an HTTP header block. Stores all headers in their original case format,
  * but allows for case-insensitive header lookup.
  */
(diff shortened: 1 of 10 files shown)


source/http_connection.c
```diff
@@ -20,6 +20,116 @@
...
+            AWS_NAPI_ENSURE(env, napi_throw_type_error(env, NULL, "Unable to convert auth_username to string"));
+            goto cleanup;
+        }
+    }
+
+    napi_value node_password = *arg++;
+    if (!aws_napi_is_null_or_undefined(env, node_password)) {
+        binding->auth_password = aws_string_new_from_napi(env, node_password);
+        if (!binding->auth_password) {
+            AWS_NAPI_ENSURE(env, napi_throw_type_error(env, NULL, "Unable to convert auth_password to string"));
```
justinboswell: Why is this type error ENSURE'd but none of the other ones are? I don't think it matters if the throw fails, nothing you can do about it, and the next call will clean it up.

lib/common/http.ts
```diff
@@ -140,6 +140,24 @@ export class HttpHeaders {
     }
 }
 
+export enum HttpProxyAuthenticationType
+{
+    None = 0,
+    Basic = 2,
```
JonathanHenson: pretty sure this should be a 1?

PR #45, opened 2019-11-07
Implement binding.ts declaration file

ColdenCullen:

### DIFF
--- a/lib/native/binding.d.ts
+++ b/lib/native/binding.d.ts
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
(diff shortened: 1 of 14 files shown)


lib/native/binding.d.ts
```diff
@@ -0,0 +1,181 @@
...
+
+/* IO */
+export declare function error_code_to_string(error_code: number): string;
+export declare function error_code_to_name(error_code: number): string;
+export declare function io_logging_enable(log_level: number): void;
+export declare function is_alpn_available(): boolean;
+export declare function io_client_bootstrap_new(): NativeHandle;
+export declare function io_tls_ctx_new(
+    min_tls_version: number,
+    ca_filepath?: StringLike,
```
justinboswell: I don't think that's gonna work out for us, considering that we will only read it as a string on the other side for most of these. I think we just need to use string here.

Like, I challenge you to feed a ```Float64Array``` to ```ca_filepath``` :)

ColdenCullen: I wrote these to match the C implementation, not the intent. Anywhere you see `StringLike`, the C impl is calling `aws_byte_buf_init_from_napi`, and `StringLike` is the list of parameters that that function accepts.

PR #46, opened 2019-11-12
Bind Websockets, Auth & Proxies

ColdenCullen:
Binds lots of cooleo stuff to NodeLand:tm:.


source/auth.c
```diff
@@ -0,0 +1,781 @@
...
+static napi_value s_creds_provider_constructor(
+    napi_env env,
+    void *self,
+    const struct aws_napi_argument args[static AWS_NAPI_METHOD_MAX_ARGS],
+    size_t num_args) {
+
+    AWS_FATAL_ASSERT(num_args == 1);
+
+    struct aws_allocator *allocator = aws_default_allocator();
+    napi_value node_self = self;
```
justinboswell: We use ```node_this``` elsewhere in this CRT.

source/auth.c
```diff
@@ -0,0 +1,781 @@
...
+
+    struct signing_config_binding *binding = NULL;
+    struct aws_allocator *allocator = aws_default_allocator();
+
+    AWS_NAPI_CALL(env, napi_unwrap(env, js_object, (void **)&binding), {
+        napi_throw_error(env, NULL, "Failed to unwrap aws_signing_config_aws");
+        return NULL;
+    });
+
+    /* Copy the node_param_blacklist into native memory */
```
justinboswell: LIES

source/auth.c
```diff
@@ -0,0 +1,781 @@
...
+struct signing_config_binding {
+    struct aws_signing_config_aws base;
+
+    struct aws_byte_buf region;
+    struct aws_byte_buf service;
+
+    napi_ref date;
+    napi_ref node_param_blacklist;
+
+    struct aws_array_list param_blacklist; /* aws_string * */
```
justinboswell: Maybe document why this is here, to allow synchronous flow

source/auth.c
```diff
@@ -0,0 +1,781 @@
...
+    size_t num_args) {
+
+    struct aws_allocator *allocator = aws_default_allocator();
+
+    struct signing_config_binding *binding = aws_mem_calloc(allocator, 1, sizeof(struct signing_config_binding));
+    binding->base.config_type = AWS_SIGNING_CONFIG_AWS;
+    binding->base.should_sign_param = s_should_sign_param;
+    binding->base.should_sign_param_ud = binding;
+
+    if (num_args >= 1 && args[0].type == napi_number) {
```
justinboswell: Do the ptr advancing thing instead, so Mike doesn't kill us

source/class_binder.c
```diff
@@ -0,0 +1,431 @@
...
+    const struct aws_napi_method_info *ctor_method;
+
+    napi_ref constructor;
+    bool is_wrapping;
+};
+
+/* Make sure our static storage is big enough */
+AWS_STATIC_ASSERT(sizeof(struct aws_napi_class_info) >= sizeof(struct aws_napi_class_info_impl));
+
+/**
```
justinboswell: This doesn't need **, it's not public

source/class_binder.c
```diff
@@ -0,0 +1,431 @@
...
+    /* Check if we're wrapping an existing object or creating a new one */
+    if (clazz->is_wrapping) {
+        AWS_FATAL_ASSERT(num_args == 1);
+
+        void *native = NULL;
+
+        /* Arg 1 should be an external */
+        AWS_NAPI_ENSURE(env, napi_get_value_external(env, node_args[0], &native));
+
+        /* Wrap shouldn't take a finalizer, because it's very likely that this object isn't owned by JS */
```
justinboswell: Note that the external has the finalizer

source/class_binder.c
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    AWS_NAPI_CALL(env, napi_unwrap(env, node_this, &self), {
+        napi_throw_error(env, NULL, "Class binder property getter must be called on a wrapped object");
+        return NULL;
+    });
+
+    const struct aws_napi_property_info *property = data;
+
+    napi_value result = property->getter(env, self);
```
justinboswell: ```AWS_FATAL_ASSERT(property->getter)```?

source/class_binder.c
```diff
@@ -0,0 +1,431 @@
...
+    });
+
+    const struct aws_napi_property_info *property = data;
+
+    struct aws_napi_argument new_value;
+    if (s_argument_parse(env, node_value, property->type, false, &new_value)) {
+        return NULL;
+    }
+
+    property->setter(env, self, &new_value);
```
justinboswell: ```AWS_FATAL_ASSERT(property->setter)```?

source/class_binder.c
```diff
@@ -0,0 +1,431 @@
...
+        napi_throw_error(env, NULL, "Failed to dereference constructor value");
+        return status;
+    });
+
+    impl->is_wrapping = true;
+    AWS_NAPI_CALL(env, napi_new_instance(env, constructor, 1, &to_wrap, result), {
+        napi_throw_error(env, NULL, "Failed to construct class-bound object");
+        return status;
+    });
+    impl->is_wrapping = false;
```
justinboswell: Comment that the ctor will call ```napi_wrap```

source/http_message.c
```diff
@@ -0,0 +1,468 @@
...
+            .setter = s_path_set,
+            .attributes = napi_enumerable | napi_writable,
+        },
+        {
+            .name = "body",
+            .setter = s_body_set,
+            .attributes = napi_enumerable | napi_writable,
+        },
+        {
+            .name = "num_headers",
```
justinboswell: Should this be like ```.size``` or ```.length```?

test/auth.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    expect(config.should_normalize_uri_path).toBe(true);
+
+    expect(config.sign_body).toBe(true);
+});
+
+const DATE_STR = '2015-08-30T12:36:00Z';
+
+// Test values copied from aws-c-auth/tests/aws-sig-v4-test-suite/get-vanilla"
+const SIGV4TEST_ACCESS_KEY_ID = 'AKIDEXAMPLE';
```
justinboswell: You should probably have a test that compares our results against the AWS SDK for JS, we already have a dependency on it. Feel free to do that as a follow up, but we should have it so we can prove correctness in the future when something breaks.

source/auth.c
ColdenCullen: not a lies

source/class_binder.c
ColdenCullen: OOPS this is phat wrong

source/http_message.c
ColdenCullen: Will revisit when Mike finishes his Headers class, but for now, request.length isn't really clear enough.

test/auth.ts
ColdenCullen: Will do as follow up :)

test/auth.ts
```diff
@@ -0,0 +1,114 @@
...
+        native.AwsSigningAlgorithm.SigV4QueryParam,
+        provider,
+        'us-east-1',
+        'iotcore',
+        now,
+        ["abc"],
+        true,
+        true,
+        true,
+    );
```
graebm: I thought you were going to do something like:
```
let config = new native.AwsSigningConfig({
    algorithm: native.AwsSigningAlgorithm,
    provider: provider,
    region: 'us-east-1',
    ...
});
```
so the 9 args were named and users could leave them out to get defaults?

lib/browser/http.ts
```diff
@@ -13,13 +13,27 @@
...
+export class HttpRequest {
+    constructor(
+        /** The verb to use for the request (i.e. GET, POST, PUT, DELETE, HEAD) */
+        public method: string,
+        /** The URI of the request */
+        public path: string,
+        /** The request body, in the case of a POST or PUT request */
+        public body?: InputStream,
+        /** Additional custom headers to send to the server */
+        public headers = new HttpHeaders()) {
```
graebm: [sync] in aws-crt-python, we're taking name-value-pairs here, not an actual HttpHeaders instance

test/auth.ts
```diff
@@ -0,0 +1,114 @@
...
+    expect(config.use_double_uri_encode).toBe(true);
+
+    expect(config.should_normalize_uri_path).toBe(true);
+
+    expect(config.sign_body).toBe(true);
+});
+
+const DATE_STR = '2015-08-30T12:36:00Z';
+
+// Test values copied from aws-c-auth/tests/aws-sig-v4-test-suite/get-vanilla"
```
graebm: this comment isn't accurate anymore.
I have no idea where you got this Authorization signature from. I searched for the string in aws-c-auth and don't see where it came from.

graebm: Ah!
I got the python test results matching those from aws-c-auth!

You need to set sign_body=false.
The "x-amz-content-sha256" header will still get added, but it won't be accounted for in the "Authorization" signature.  I made the python test less strict, checking that the signed headers were PRESENT, but not don't fail if there are extra headers present

lib/native/auth.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ClientBootstrap } from './io';
+
+export enum AwsSigningAlgorithm {
+    SigV4Header,
+    SigV4QueryParam,
+}
+
+/* Subclass for the purpose of exposing a non-NativeHandle based API */
+export class AwsCredentialsProvider extends crt_native.AwsCredentialsProvider {
+    static newDefault(bootstrap: ClientBootstrap): AwsCredentialsProvider {
```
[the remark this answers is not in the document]
graebm: Just changed python to go the factory-function route.
Only difference now is python uses term "new default chain" and JS uses "new default"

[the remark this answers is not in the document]
justinboswell: Let's pick a pattern and be consistent across the CRTs. What does C++ do?

lib/browser/mqtt.ts
```diff
@@ -102,21 +102,16 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
-        return WebsocketUtils.transform_websocket_url(url, this.config);
-    }
-
     constructor(
         readonly client: MqttClient,
         private config: MqttConnectionConfig) {
         super();
 
+        const create_websocket_stream = (client: MqttClientInternal) => WebsocketUtils.create_websocket_stream(this.config);
+        const transform_websocket_url = (url: string, options: IClientOptions, client: MqttClientInternal) => WebsocketUtils.create_websocket_url(this.config);
```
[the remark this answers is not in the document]
ColdenCullen: So the thing about Javascript is that `async` isn't part of the signature, so it's not really required here. Also I didn't change any functionality here, it's just a quick refactor to better scope stuff.

lib/common/http.ts
```diff
@@ -61,10 +84,22 @@ export class HttpHeaders {
      * @return List of values, or empty list if none exist
      */
     get_values(name: string) {
-        const values_list = this.headers[name.toLowerCase()] || [];
+        const key = name.toLowerCase();
+
         const values = [];
```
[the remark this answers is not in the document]
ColdenCullen: We already have that here, see `[Symbol.iterator]`. Should I remove this function? @justinboswell  wrote this originally, do you have thoughts?

lib/common/io.ts
```diff
@@ -32,6 +32,3 @@ export enum SocketDomain {
     IPV6 = 1,
     LOCAL = 2, /* UNIX domain/named pipes */
 }
-
-/** Base class for streams used as inputs to AWS CRT I/O facilities */
-export interface InputStreamBase { };
```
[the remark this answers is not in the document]
ColdenCullen: Resolved offline, closing

lib/native/auth.ts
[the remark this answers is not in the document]
ColdenCullen: Factory is the pattern! You guys picked mine while I was out, so thanks

lib/browser/http.ts
ColdenCullen: We decided to keep it going this way, as it's how Java also works.

test/auth.ts
ColdenCullen: Will revisit once Bret finishes unfactoring signing.

lib/native/auth.ts
```diff
@@ -29,43 +29,20 @@ export class AwsCredentialsProvider extends crt_native.AwsCredentialsProvider {
...
-    }
-}
+export type AwsSigningConfig = crt_native.AwsSigningConfig;
 
-export class AwsSigner extends crt_native.AwsSigner {
-
-    public async sign_request(request: HttpRequest, config: AwsSigningConfig): Promise<HttpRequest> {
-        return new Promise((resolve, reject) => {
-            super.sign_request(request, config, (error_code) => {
+export async function sign_request_aws(request: HttpRequest, config: AwsSigningConfig): Promise<HttpRequest> {
```
graebm: naming: sign_aws_request() ???

graebm: aws_sign_request() ???

graebm: I like aws_sign_request(). everything else in this file is prefixed "Aws" and it sounds good

PR #56, opened 2019-12-04
Release Pipeline

ColdenCullen:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "aws-crt",
-  "version": "0.4.1",
+  "version": "0.4.3",
```
graebm: Is there a way to formally state the requirement for node v10.16+ here in the package file?

ColdenCullen: Not without breaking browser compatibility :(

PR #66, opened 2020-03-12
Make linux use an older glibc

justinboswell:
*Issue #, if available:* https://github.com/aws/aws-iot-device-sdk-js-v2/issues/26

*Description of changes:* Updated dependencies and switched linux binary build to be on AL2012 to get an old enough glibc (2.17)

As per https://gist.github.com/wagenet/35adca1a032cec2999d47b6c40aa45b1 we should be good with 2.17 on all distros we intend to support.

### DIFF
--- a/continuous-delivery/build-binaries-linux-x64.yml
+++ b/continuous-delivery/build-binaries-linux-x64.yml
@@ -4,6 +4,7 @@ phases:
     commands:
       - mkdir linux-x64
       - cd aws-crt-nodejs
+      - env
       - ls -R
       - builder build --project=.
       - cp -r dist/bin/linux-x64/* ../linux-x64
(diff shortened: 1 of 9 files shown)


continuous-delivery/build-binaries-linux-x64.yml
```diff
@@ -1,16 +1,11 @@
...
-      nodejs: 10
-    commands:
-      - apt-get install -y cmake
   build:
     commands:
       - mkdir linux-x64
       - cd aws-crt-nodejs
-      - bash ./continuous-delivery/build-binaries-unix.sh
-      - cp -r dist/bin/linux-x64/* ../linux-x64
+      - builder --version=v0.6.0 build --project=aws-crt-nodejs run_tests=false
```
bretambrose: Is this going to create version sync problems?  Any way they can stay together?

justinboswell: good catch, that was from debugging, will fix.

PR #72, opened 2020-03-31
Updated submodules to latest stable versions, fixed tests, added new …

JonathanHenson:
…stream unactive test. Elasticurl is hopelessly broken and unreadable and should be rewritten.


source/mqtt_client_connection.c
```diff
@@ -72,8 +71,9 @@ napi_value aws_napi_mqtt_client_connection_close(napi_env env, napi_callback_inf
         return NULL;
     });
 
-    /* no more node interop will be done, free node resources */
```
justinboswell: Probably worth keeping this comment

PR #90, opened 2020-05-29
Added jest-puppeteer support for browser tests

justinboswell:
* Unified build script (install.js and build.js are no longer copy pasta)
* Added -D argument support to build script:
  * ```npm run install -- -DLibCrypto_INCLUDE_DIR=/opt/ssl/include```
* Moved tests with their impls
* Added ```test:node``` and ```test:browser``` targets. ```test``` command/target runs the node tests only by default. ```test:browser``` will install puppeteer and prereqs first.


justinboswell:
Example run of browser tests: https://github.com/awslabs/aws-crt-nodejs/pull/90/checks?check_run_id=719021654#step:2:1500

lib/common/platform.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import * as platform from './platform'
+
+test('platform.is_nodejs is correct', () => {
```
justinboswell, line 18: Just some basic tests that run on both platforms

test/native/jest.config.js
```diff
@@ -11,12 +14,12 @@ module.exports = {
...
+        "^.+\\.ts?$": ['ts-jest'],
     },
     testPathIgnorePatterns: [
         '/node_modules/'
     ],
     modulePathIgnorePatterns: [
         '/cmake-js/'
     ],
-    testRunner: "jest-circus/runner"
 }
```
graebm, line 25: trivial: lots of these files have no newline at end of file.

justinboswell: The JS linter actually removes them

justinboswell:
> Saw this at the end of the output. If async ops hanging past the end of a run is a wild west situation, it may not be critical, but the output may be nice to have:
> 
> > Force exiting Jest: Have you considered using `--detectOpenHandles` to detect async operations that kept running after all tests finished?
> 
> Looking at ci.yml, I'm not seeing how Windows is being excluded from the browser tests. It makes sense that they can't run on Windows though.

The browser tests only run on al2, because that's the only place npm run test:browser:ci is run.

As for the open handles thing, one day we will figure out why that happens, but jest is the only thing that ever reports it as a problem. Normal JS programs don't have any issue exiting.

PR #92, opened 2020-05-29
Generalized more tests to be common

justinboswell:


lib/browser/mqtt.ts
```diff
@@ -12,8 +12,8 @@
 * permissions and limitations under the License.
 */
 
-import { MqttClient as MqttClientInternal, IClientOptions, ISubscriptionGrant, IUnsubackPacket, IPublishPacket } from "mqtt";
-import { AsyncClient } from "async-mqtt";
+import { AsyncClient, IClientOptions, ISubscriptionGrant, IUnsubackPacket, IPublishPacket, IConnackPacket } from "async-mqtt";
```
justinboswell, line 15: In a follow up, we can get rid of async-mqtt, import mqtt directly, and just do the same wrapping async-mqtt does, it's not special.

lib/browser/http.ts
```diff
@@ -173,15 +175,15 @@ export class HttpRequest {
...
-        scheme?: string,
+        protected socket_options: SocketOptions,
+        protected tls_opts?: TlsConnectionOptions,
         proxy_options?: HttpProxyOptions,
     ) {
         super();
-        if (!scheme) {
-            scheme = (port == 443) ? 'https' : 'http'
-        }
+        let scheme = (tls_opts || port == 443) ? 'https' : 'http'
```
JonathanHenson, line 186: what if I'm using a non-standard tls port? like 8883 or 8443?

lib/native/mqtt.spec.ts
```diff
@@ -221,7 +218,7 @@ test('MQTT Websocket', async () => {
 
 test('MQTT Pub/Sub', async () => {
 
-    io.enable_logging(io.LogLevel.TRACE);
+    //io.enable_logging(io.LogLevel.TRACE);
```
JonathanHenson, line 221: you want to remove this after testing?

test/native/jest.config.js
```diff
@@ -22,4 +22,10 @@ module.exports = {
     modulePathIgnorePatterns: [
         '/cmake-js/'
     ],
+    setupFiles: ['<rootDir>/test/native/jest.setup.ts'],
+    moduleNameMapper: {
+        '@common/(.+)': '<rootDir>/lib/common/$1',
+        '@awscrt/(.+)': '<rootDir>/lib/native/$1',
+        '@awscrt': '<rootDir>/lib/index.ts'
+    }
 }
```
JonathanHenson, line 31: newline?

tsconfig.json
```diff
@@ -5,5 +5,15 @@
...
+        "paths": {
+            "@awscrt": [
+                "lib/index.ts"
+            ],
+            "@awscrt/*": [
+                "lib/native/*"
+            ]
+        }
+    }
 }
```
JonathanHenson, line 19: newline?

lib/browser/http.ts
justinboswell: yeah, I guess we could just do port != 80? But then, what about 8080? Maybe we shouldn't try to interpret port?

lib/native/mqtt.spec.ts
justinboswell: I'll delete it, it's left from some testing a while back.

lib/browser/http.ts
```diff
@@ -236,6 +238,10 @@ export class HttpClientConnection extends BufferedEventEmitter {
     _on_end(stream: HttpClientStream) {
         this.emit('close');
     }
+
+    close() {
+        this.emit('close');
+    }
```
graebm, line 244: This comment doesn't apply to this spot, I'm juts digging into this impl for the first time. I see that it's hooked up so we "end" the connection after it handles 1 response. In the C impl, this would be like an HTTP/1.0 response coming back, or a "connection: close" header happening. In the C impl, we take care that the connection.is_open() returns false during the stream's "end" callback, so that if we hand it back to a connection pool from that callback, the pool knows that the connection is not to be reused.

anyway, we might want to also be sure that `this.axios = null` before invoking the stream's "end" event.

This is a lot of design overhaul feedback for an "improve tests" PR, so we might want to address some of the design differences in a future PR

graebm: Looks like axios has the ability to "cancel" requests, we might want to use that to cancel requests when the user calls connection.close(), and hold off with the connection's close/end event until those streams finish wrapping up

lib/browser/http.ts
```diff
@@ -287,6 +293,18 @@ export class HttpClientStream extends BufferedEventEmitter {
...
 
+    /**
+     * Begin sending the request.
+     *
+     * The stream does nothing until this is called. Call activate() when you
+     * are ready for its callbacks and events to fire.
+     */
+    activate() {
+        setTimeout(() => {
+            this.uncork();
```
graebm, line 304: I don't see `this.cork()` being called anywhere?

lib/browser/http.ts
```diff
@@ -173,15 +175,15 @@ export class HttpRequest {
 export class HttpClientConnection extends BufferedEventEmitter {
     readonly axios: any;
     constructor(
+        protected boostrap: ClientBootstrap,
```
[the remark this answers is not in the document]
graebm: more importantly.
It's cool that we're passing ClientBootstrap into the browser API so everything's compatible,
but maybe we want to make stuff like this typescript-optional, and in browser/ we ignore it either way, but in native/ we actually raise an exception if it's undefined?

That would let people/tests use the same exact code, but if you were writing browser-only code you could skip some of the unnecessary song and dance

lib/browser/mqtt.ts
```diff
@@ -291,26 +301,12 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
-                this.connection.on('end', this.on_disconnected);
-            } catch (e) {
-                reject(e);
-            }
+            this.connection.once('connect', (connack: IConnackPacket) => {
+                resolve(connack.sessionPresent);
+            });
+            this.connection.once('error', (error: Error) => {
+                reject(new CrtError(error));
+            });
```
graebm, line 309: just FYI, when things weren't working on my machine (and I never figured out 100% why), I was seeing a "close" event from this.connection instead of an "error" event
not sure if we need to tackle that right now, just saying...

package.json
```diff
@@ -43,7 +43,7 @@
     "async-mqtt": "^2.5.0",
```
graebm, line 43: might want to bump async-mqtt to 2.6.0
asnc-mqtt 2.5.0 is still using mqtt 2.3.1: https://github.com/mqttjs/async-mqtt/blob/v2.5.0/package.json#L28

PR #95, opened 2020-06-05
Generalized/corrected MQTT tests, removed async-mqtt dependency

justinboswell:
* MQTT tests now work both in browser and in node


lib/browser/aws_iot.ts
```diff
@@ -42,6 +42,14 @@ export class AwsIotMqttConnectionConfigBuilder {
         };
     }
 
+    static new_mtls_builder(...args: any[]) {
```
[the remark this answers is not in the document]
graebm: Make that comment in the code too

lib/browser/mqtt.ts
```diff
@@ -300,13 +295,16 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
-            this.connection.once('connect', (connack: IConnackPacket) => {
-                resolve(connack.sessionPresent);
-            });
-            this.connection.once('error', (error: Error) => {
+            const on_connect_error = (error: Error) => {
                 reject(new CrtError(error));
+            };
+            this.connection.once('connect', (connack: mqtt.IConnackPacket) => {
+                this.connection.removeListener('error', on_connect_error);
+                resolve(connack.sessionPresent);
```
graebm, line 305: Just curious why you added the `.removeListener()` call.
[Docs](https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Promise.jsm/Promise) say resolve()/reject() do nothing if the promise was already fulfilled.
It is cleaner, but not strictly necessary.

lib/browser/mqtt.ts
```diff
@@ -337,13 +335,15 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
-            .catch((reason) => {
-                this.emit('error', new CrtError(reason));
-            })
-            .then((value) => {
-                return { packet_id: (value as IPublishPacket).messageId };
+        return new Promise((resolve, reject) => {
+            this.connection.publish(topic, payload_data, { qos: qos, retain: retain }, (error, packet) => {
+                if (error) {
+                    reject(new CrtError(error));
+                    return this.on_error(error);
```
graebm, line 342: why `return` here? on_error doesn't return anything, and this is a callback where the caller doesn't expect a return value

Are you doing this just to early-out? If that's the case, a lone `return;` would be less confusing. or just use an else{} block

lib/browser/mqtt.ts
```diff
@@ -367,14 +367,16 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
-                this.emit('error', new CrtError(reason));
-            })
-            .then((value) => {
-                const sub = (value as ISubscriptionGrant[])[0];
-                return { topic: sub.topic, qos: sub.qos };
+        return new Promise((resolve, reject) => {
+            this.connection.subscribe(topic, { qos: qos }, (error, packet) => {
+                if (error) {
+                    reject(new CrtError(error))
+                    return this.on_error(error);
```
graebm, line 374: similar, what's with the `return`?

lib/browser/ws.ts
```diff
@@ -85,7 +85,7 @@ function sign_url(method: string,
     if (credentials.aws_sts_token) {
         query_params += `&X-Amz-Security-Token=${encodeURIComponent(credentials.aws_sts_token)}`;
     }
-    const signed_url = `${url.protocol}${url.hostname}${url.pathname}${query_params}`;
+    const signed_url = `${url.protocol}//${url.hostname}${url.pathname}${query_params}`;
```
graebm, line 88: 😬

lib/native/mqtt.spec.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
graebm, line 2: 👩‍⚖️ we can do this? I am so on board if so

package.json
```diff
@@ -27,22 +27,25 @@
...
     "jest-runtime": "^25.2.4",
+    "puppeteer": "^3.3.0",
     "ts-jest": "^25.3.0",
     "typedoc": "^0.17.7",
     "typescript": "^3.8.3",
     "uuid": "^3.4.0",
     "yargs": "^13.3.2"
   },
   "dependencies": {
-    "async-mqtt": "^2.5.0",
```
graebm, line 43: 👋

lib/browser/mqtt.ts
justinboswell: I added the removeListener() call because otherwise, the first time there's an error, this callback will trigger.

lib/browser/ws.ts
justinboswell: Tell me about it.

lib/browser/mqtt.ts
justinboswell: Early out.

lib/native/mqtt.spec.ts
justinboswell: Yes. It's a soft push across the org, but I have a task in the Fix-It Friday backlog to delete years from our copyrights.

lib/browser/mqtt.ts
justinboswell: This is what TS/JS people do 🤷‍♂️ . It's just an early out.

lib/browser/mqtt.ts
graebm: ohhhhhhh yeah
👍

lib/browser/mqtt.ts
graebm: And here I was, with all this respect for JS now that I've learned how cool the modern ecosystem is

and now it's all gone

j/k ok whatever

lib/browser/mqtt.ts
```diff
@@ -367,14 +367,16 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
+                }
+                const sub = (packet as mqtt.ISubscriptionGrant[])[0];
+                resolve({ topic: sub.topic, qos: sub.qos });
```
graebm, line 377: Looking a little harder since you asked about failing tests...
if we get back sub.qos==128, that means the subscription failed.

For single-topic subscribe(), aws-crt-python turns that into an error, instead of passing it through the QoS.

But for resubscribe_existing_topics() (which is a multi-topic subscribe under the hood), aws-crt-python passes None for the QoS to indicate that a specific topic failed.

lib/browser/polyfills.ts
```diff
@@ -0,0 +1,19 @@
+/* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
bretambrose, line 1: Minor: new license format

lib/common/http.spec.ts
```diff
@@ -1,3 +1,17 @@
+/*
```
bretambrose, line 1: license format

lib/common/mqtt.spec.ts
```diff
@@ -0,0 +1,203 @@
+/*
```
bretambrose, line 1: license format

Issue #103, opened 2020-07-17
Polyfill not included in published browser build


stevelacy:
The `polyfill.js` file is required here: https://github.com/awslabs/aws-crt-nodejs/blob/master/lib/browser/http.ts#L13
But not found in the published npm module:
```
$ tree node_modules/aws-crt/dist.browser/browser

├── aws_iot.d.ts
├── aws_iot.js
├── aws_iot.js.map
├── crypto.d.ts
├── crypto.js
├── crypto.js.map
├── error.d.ts
├── error.js
├── error.js.map
├── http.d.ts
├── http.js
├── http.js.map
├── io.d.ts
├── io.js
├── io.js.map
├── mqtt.d.ts
├── mqtt.js
├── mqtt.js.map
├── trie.d.ts
├── trie.js
├── trie.js.map
├── ws.d.ts
├── ws.js
└── ws.js.map
```

Doff3n:
I have the same problem. Only .ts files are included in tsconfig.browser.json.

If I manually include polyfill.js I also find that Create React app cannot resolve @awscrt/io.

stevelacy:
@Doff3n Yeah found that issue as well, quick hack is to remove the `@awscrt/io` and replace it with `../io` then run tsc

justinboswell:
This has been fixed by the latest commit. We will roll a new version ASAP.

PR #112, opened 2020-07-22
Remove tsc @awscrt alias preventing browser lib from running

stevelacy:
This fixes an issue where the typescript build alias is preserved after compile. Node or a browser is unable to resolve the destination of the alias, and thus throws a module not found error.
I did not find any other occurrences of the aliases used in the ts project. 

This issue is referenced in this comment: https://github.com/awslabs/aws-crt-nodejs/issues/103#issuecomment-662358877


pickfire:
It is broken for me too, hope this fix it.

bretambrose:
Thanks for pointing this out.  I folded it into a larger set of fixes for the browser shim and samples.

PR #118, opened 2020-08-13
Browser fixes

bretambrose:
A grab bag of browser shim fixes:

* Update http sample to use new http interface
* Webpack error in http module
* Corked event args not unpacking properly leading to undefined arguments
* mqtt connect timeout default should be a reasonable value in milliseconds, not seconds
* activate http streams in browser http sample
* rework HttpClientConnection constructor to not be confusing and erroneously ambiguous with null/undefined parameters
* audit fix for vulnerable dependencies

### DIFF
--- a/lib/browser/http.ts
+++ b/lib/browser/http.ts
@@ -168,30 +168,23 @@ export class HttpClientConnection extends BufferedEventEmitter {
     protected socket_options?: SocketOptions;
     protected tls_options?: TlsConnectionOptions;
     protected proxy_options?: HttpProxyOptions;
-
-    /**
-     * Browser-specific overload of constructor without bootstrap
-     */
-    constructor(host_name: string, port: number, socket_options?: SocketOptions, tls_options?: TlsConnectionOptions, proxy_options?: HttpProxyOptions);
-    constructor(bootstrap: ClientBootstrap, host_name: string, port: number, socket_options?: SocketOptions, tls_options?: TlsConnectionOptions, proxy_options?: HttpProxyOptions);
+    
     constructor(
-        bootstrapOrHost: ClientBootstrap | string,
-        hostOrPort: string | number,
-        portOrSocketOptions?: number | SocketOptions,
-        socketOptionsOrTlsOptions?: SocketOptions | TlsConnectionOptions,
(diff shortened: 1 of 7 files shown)


lib/browser/http.ts
```diff
@@ -356,7 +349,6 @@ export class HttpClientStream extends BufferedEventEmitter {
         }
         this.emit('data', data);
         this.emit('end');
-        this.connection.close();
```
bretambrose, line 359: Only change I'm not certain about.  Neither was Justin.

lib/browser/http.ts
```diff
@@ -169,29 +169,22 @@ export class HttpClientConnection extends BufferedEventEmitter {
...
-    constructor(host_name: string, port: number, socket_options?: SocketOptions, tls_options?: TlsConnectionOptions, proxy_options?: HttpProxyOptions);
-    constructor(bootstrap: ClientBootstrap, host_name: string, port: number, socket_options?: SocketOptions, tls_options?: TlsConnectionOptions, proxy_options?: HttpProxyOptions);
     constructor(
-        bootstrapOrHost: ClientBootstrap | string,
-        hostOrPort: string | number,
-        portOrSocketOptions?: number | SocketOptions,
-        socketOptionsOrTlsOptions?: SocketOptions | TlsConnectionOptions,
-        tlsOptionsOrProxyOptions?: TlsConnectionOptions | HttpProxyOptions,
-        maybeProxyOptions?: HttpProxyOptions,
+        bootstrap: ClientBootstrap,
```
justinboswell, line 173: Any reason not to make this ```| undefined```?

lib/common/event.ts
```diff
@@ -19,7 +19,7 @@ type EventKey = string | symbol;
 class BufferedEvent {
     public next?: BufferedEvent;
     public args: any[];
-    constructor(public event: EventKey, ...args: any[]) {
+    constructor(public event: EventKey, args: any[]) {
```
justinboswell, line 22: OH NO I'M SO SORRY

lib/browser/http.ts
bretambrose: Seems like it would be ok, will give it a try

PR #126, opened 2020-08-25
Request body

TingDaoK:
*Issue #, if available:*

HttpRequest is not taking InputStream, but the native_handler as argument.

*Description of changes:*

- Added a http.HttpRequest derived from crt_native.HttpRequest
- Fix the refcounts for request

*TODO:*
enable elasticurl from CI will be helpful to uncover this kind of bug


samples/node/elasticurl/elasticurl.ts
```diff
@@ -230,7 +230,8 @@ async function main(argv: Args){
         }
 
         return new Promise((resolve, reject) => {
-            const request = new http.HttpRequest(argv.method, argv.url.toString(), headers, body_stream);
+            let body_stream_handler = body_stream ? body_stream.native_handle():undefined;
```
TingDaoK, line 233: Hmm, probably should change the constructor instead.

source/http_stream.c
```diff
@@ -192,8 +192,8 @@ static void s_http_stream_binding_finalize(napi_env env, void *finalize_data, vo
     (void)finalize_hint;
     struct http_stream_binding *binding = finalize_data;
 
-    aws_http_message_destroy(binding->request);
-    aws_http_message_destroy(binding->response);
+    aws_http_message_release(binding->request);
```
TingDaoK, line 195: Actually `aws_http_message_destroy ` just call `aws_http_message_release `. But I just feel it's more clear to call `aws_http_message_release `

PR #129, opened 2020-09-03
Upgrading bl dependency

DavidOgunsAWS:
*Issue #, if available:*
 * Update requested

*Description of changes:*
 * Dependency update of one dependency


bretambrose:
Why did it need to be added though?  Doesn't npm audit just fix the package-lock.json for us even for deeply downstream dependencies?

package.json
```diff
@@ -31,6 +31,7 @@
     "@types/uuid": "^3.4.8",
     "@types/ws": "^6.0.4",
     "aws-sdk": "^2.640.0",
+    "bl": "^2.2.1",
```
graebm, line 34: comment that we don't need this directly, but we want this version for security

DavidOgunsAWS: Removed, and then let npm update deal with all dependency updates in package-lock.json for now.

graebm (approved):

PR #131, opened 2020-09-08
unittest for signing body and crtError

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/native/auth.spec.ts
+++ b/lib/native/auth.spec.ts
@@ -5,6 +5,8 @@
 
 import { auth as native, http as native_http } from '../index';
 import { aws_sign_request } from './auth';
+import { InputStream } from './io';
+import { PassThrough } from "stream";
 
 const DATE_STR = '2015-08-30T12:36:00Z';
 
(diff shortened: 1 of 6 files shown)


lib/native/auth.spec.ts
```diff
@@ -57,3 +59,41 @@ test('AWS Signer SigV4 Headers', async () => {
...
+
+    const signing_config: native.AwsSigningConfig = {
+        algorithm: native.AwsSigningAlgorithm.SigV4,
+	    signature_type: native.AwsSignatureType.HttpRequestViaHeaders,
+        provider: credentials_provider,
+        region: SIGV4TEST_REGION,
+        service: SIGV4TEST_SERVICE,
+        date: new Date(DATE_STR),
+        // signed_body_value: native.AwsSignedBodyValue.UnsignedPayload,
+        signed_body_header: native.AwsSignedBodyHeaderType.None,
```
TingDaoK, line 78: As long as the signed_body_value is not set, we will need to read from the body for the `CanonicalRequest`. And it will lead to an endless loop for reading from input stream.

lib/native/auth.ts
```diff
@@ -128,11 +129,18 @@ export type AwsSigningConfig = crt_native.AwsSigningConfig;
 export async function aws_sign_request(request: HttpRequest, config: AwsSigningConfig): Promise<HttpRequest> {
     return new Promise((resolve, reject) => {
         try {
+            if(!request.bodyStream?.isCompleted()) {
+                /* The body has not been fully loaded, we cannot sign the body. Attempt to send the request 
+                 * with an unsigned payload, which may or may not be accepted by the service. */
```
TingDaoK, line 134: Same approach as JS SDK.

justinboswell (approved):

lib/native/auth.ts
```diff
@@ -128,11 +129,13 @@ export type AwsSigningConfig = crt_native.AwsSigningConfig;
 export async function aws_sign_request(request: HttpRequest, config: AwsSigningConfig): Promise<HttpRequest> {
     return new Promise((resolve, reject) => {
         try {
+            /* TODO: if the body of request has not fully loaded, it will lead to a endless loop. 
+             * User should set the config to prevent this endless loop for now */
```
graebm, line 133: trivial: delete? It's a "TODO", but I'm not sure what this is saying we should do?

lib/native/auth.spec.ts
```diff
@@ -5,6 +5,8 @@
 
 import { auth as native, http as native_http } from '../index';
 import { aws_sign_request } from './auth';
+import { InputStream } from './io';
+import { PassThrough } from "stream";
```
graebm, line 9: trivial: still necessary? These are the only changes in the file

graebm (approved):

lib/native/auth.spec.ts
TingDaoK: They are used by the tests

lib/native/auth.ts
TingDaoK: Yeah... It's a little bit weird... The idea is to make some notes for future reference to avoid the endless loop... May make it as a Note or something?

Issue #79, opened 2020-05-01
Unclear reconnection handling with reconnectPeriod: 0 in mqtt options


burritoIand:
Hello,

I noticed that the default value for the underlying mqtt client.options.reconnectPeriod is `0`. Upon looking through the mqtt library itself I discovered that a value of 0 means the client will flush at the end and _not_ try to reconnect. Is this accurate and expected?

https://github.com/mqttjs/MQTT.js/issues/1083

PR #133, opened 2020-09-14
add reconnect_period param

RaulLazaro:
*Issue #, if available:*
#79

*Description of changes:*
Add the possibility to change the reconnect period, also enabling the ability to reconnect.
The default value of 0 is disabling reconnection.

There is a console error for each reconnection attempt that I have not been able to eliminate
`WebSocket connection to 'wss://...' failed: Error in connection establishment: net::ERR_INTERNET_DISCONNECTED`


edzis:
It is also not clear to me how should reconnection happen. From the workflow described in https://github.com/mqttjs/MQTT.js#about-reconnection this lib seems to pass `transformWsUrl` but hardcodes the `reconnectPeriod`, preventing any usage of the `transformWsUrl` as far as I can see.

PR #139, opened 2020-09-28
fixed converting the headers

TingDaoK:
*Issue #, if available:*

- When the string is larger than 256, the `napi_get_value_string_utf8` cannot fit into the buffer with the same capacity.

*Description of changes:*

- make the capacity 1 byte larger than the length.

### DIFF
--- a/source/http_headers.c
+++ b/source/http_headers.c
@@ -237,12 +237,12 @@ static napi_value s_headers_constructor(napi_env env, const struct aws_napi_call
                 napi_throw_type_error(env, NULL, "HTTP header was not a string or length could not be extracted");
                 goto header_parse_error;
             });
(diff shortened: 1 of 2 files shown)


source/http_headers.c
```diff
@@ -237,12 +237,12 @@ static napi_value s_headers_constructor(napi_env env, const struct aws_napi_call
                 napi_throw_type_error(env, NULL, "HTTP header was not a string or length could not be extracted");
                 goto header_parse_error;
             });
-            aws_byte_buf_reserve(&name_buf, length);
+            aws_byte_buf_reserve(&name_buf, length + 1);
```
justinboswell, line 240: GOOD CATCH

justinboswell (approved):

PR #145, opened 2020-11-04
Use dynamically bound openssl api discovery

bretambrose:
* Switches the nodejs crt to use the dynamically discovered openssl branch of aws-c-cal

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -24,7 +24,7 @@ for (const arg of process.argv) {
 // Run the build
 var buildSystem = new cmake.BuildSystem({
     target: "install",
-    debug: true,
+    debug: process.argv.includes('--debug'),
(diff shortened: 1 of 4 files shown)


scripts/build.js
```diff
@@ -24,7 +24,7 @@ for (const arg of process.argv) {
 // Run the build
 var buildSystem = new cmake.BuildSystem({
     target: "install",
-    debug: process.argv.includes('--debug'),
+    debug: true,
```
TingDaoK, line 27: Is it neccessary? 🤔 Why do we need to build with debug all the time?

bretambrose: Oops, did not mean to include that

TingDaoK (approved):
LGTM

PR #168, opened 2021-01-12
Fix crash on websocket reconnect.

graebm:
Bug was that `transform_websocket_args` pointer was being destroyed after 1st use, but pointer was re-used for any future reconnects, leading to access violoations. Solution is to create a new `transform_websocket_args` each time a transformation occurs.

Also moved a LOT of logic into the internal MQTT new() call, instead of the internal connect() call. I moved things that only needed to be set once (ex: websocket transform fn). I also added a bunch of cleanup calls to the finalize() function, seems like we were leaking JS resources before... but this is my first time really digging into this code so would love another set of eyeballs on this

### DIFF
--- a/lib/native/binding.d.ts
+++ b/lib/native/binding.d.ts
@@ -118,6 +118,13 @@ export function mqtt_client_connection_new(
     client: NativeHandle,
     on_interrupted?: (error_code: number) => void,
     on_resumed?: (return_code: number, session_present: boolean) => void,
+    tls_ctx?: NativeHandle,
+    will?: { topic: StringLike, payload: String | Object | DataView, qos: number, retain: boolean },
+    username?: StringLike,
+    password?: StringLike,
+    use_websocket?: boolean,
+    proxy_options?: NativeHandle,
+    websocket_handshake_transform?: (request: any, done: (error_code?: number) => void) => void,
 ): NativeHandle;
 
 /** @internal */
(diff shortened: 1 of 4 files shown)


lib/native/binding.d.ts
```diff
@@ -118,6 +118,13 @@ export function mqtt_client_connection_new(
     client: NativeHandle,
     on_interrupted?: (error_code: number) => void,
     on_resumed?: (return_code: number, session_present: boolean) => void,
+    tls_ctx?: NativeHandle,
+    will?: { topic: StringLike, payload: String | Object | DataView, qos: number, retain: boolean },
+    username?: StringLike,
+    password?: StringLike,
+    use_websocket?: boolean,
+    proxy_options?: NativeHandle,
+    websocket_handshake_transform?: (request: any, done: (error_code?: number) => void) => void,
```
justinboswell, line 127: Upon reflection and reading this again after some time removed, it feels like these function params should be typed, shouldn't they?

source/mqtt_client_connection.c
```diff
@@ -933,7 +957,7 @@ static void s_on_suback(
     args->qos = qos;
     args->packet_id = packet_id;
 
-    AWS_NAPI_ENSURE(env, aws_napi_queue_threadsafe_function(args->on_suback, args));
+    AWS_NAPI_ENSURE(NULL, aws_napi_queue_threadsafe_function(args->on_suback, args));
```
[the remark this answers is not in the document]
graebm: fixed

lib/native/binding.d.ts
graebm: did

justinboswell (approved):

PR #175, opened 2021-02-18
fix the async call crash

TingDaoK:
*Issue #, if available:*
Probably related to issues that crash from an async callback, with stacktrace saying napi_call failed from callback. Since the env may exit before the async callback and fail all the napi_calls
- https://github.com/aws/aws-iot-device-sdk-js-v2/issues/96

The fix of previous issue leads to multiple underlying issues:
- We created an eventloop-group for node, but we never clean it up, so, we will actually result in some leak
- We are using event emitter to invoke the callback, which may lead to an asynced write operation, so, if we clean up the underlying resource right after the callback, we may crash. Instead, using the finalizer to clean up the resource.

*Description of changes:*

Refering: https://nodejs.org/api/n-api.html#n_api_asynchronous_thread_safe_function_calls

> napi_unref_threadsafe_function
> This API is used to indicate that the event loop running on the main thread may exit before func is destroyed. Similar to uv_unref it is also idempotent.

- `napi_unref_threadsafe_function` will not keep env alive, so, when the callback invoked for nodejs, the env maybe invalid, and causing a crash
- Removed the unref stuff and now we need to manually release the thread_safe function to clean it up correctly. Added it for all those thread-safe function I found.
- No need to check the env, it's supposed to be valid, if not, we just screwed it up.
- Interesting finding: If `napi_release_threadsafe_function` called from the threadsafe function, we will end up with weird crash, so, instead of release, we try to use unref for those case. (I know it's hacky and weird.) But due to my experiment, that as the env cleans itself up, the threadsafe function will be finalized as well! So, I think it's not a bad solution.
- Running elasticurl in CI

### DIFF
--- a/builder.json
+++ b/builder.json
@@ -90,6 +90,17 @@
             "npm",
             "run",
 	    "test:native"
+        ],
+        [
+            "npm",
+            "install",
+            "{project_dir}/samples/node/elasticurl/"
+        ],
+        [
+            "{python}",
+            "{project_dir}/crt/aws-c-http/integration-testing/http_client_test.py",
+            "node",
+            "{project_dir}/samples/node/elasticurl/dist/elasticurl.js"
         ]
     ]
 }
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -308,15 +308,9 @@ failed_request:
 failed_external:
 failed_callbacks:
     if (binding) {
-        if (binding->on_complete) {
-            AWS_NAPI_ENSURE(env, napi_release_threadsafe_function(binding->on_complete, napi_tsfn_abort));
-        }
-        if (binding->on_response) {
-            AWS_NAPI_ENSURE(env, napi_release_threadsafe_function(binding->on_response, napi_tsfn_abort));
-        }
-        if (binding->on_body) {
-            AWS_NAPI_ENSURE(env, napi_release_threadsafe_function(binding->on_body, napi_tsfn_abort));
-        }
+        AWS_NAPI_ENSURE(env, aws_napi_release_threadsafe_function(binding->on_complete, napi_tsfn_abort));
+        AWS_NAPI_ENSURE(env, aws_napi_release_threadsafe_function(binding->on_response, napi_tsfn_abort));
+        AWS_NAPI_ENSURE(env, aws_napi_release_threadsafe_function(binding->on_body, napi_tsfn_abort));
     }
     aws_mem_release(allocator, binding);
 failed_binding_alloc:
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -23,6 +23,8 @@ struct http_stream_binding {
     napi_threadsafe_function on_body;
     struct aws_http_message *response; /* used to buffer response headers/status code */
     struct aws_http_message *request;
+
+    uint64_t pending_length; /* used to check all the body callbacks for nodejs has been invoked */
 };
 
 static void s_on_response_call(napi_env env, napi_value on_response, void *context, void *user_data) {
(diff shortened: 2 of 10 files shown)


martenpayne (commented):
Fixes race condition I'm seeing with JS SDK v2

source/module.h
```diff
@@ -58,8 +58,8 @@ napi_status aws_napi_dispatch_threadsafe_function(
     napi_value *argv);
 
 /**
- * Wrapper around napi_create_threadsafe_function that ensures it is a weak reference and cleans
- * up when the last node reference is cleared.
+ * Wrapper around napi_create_threadsafe_function,
+ * the theadsafe function needed to be released to clean up correctly
```
justinboswell, line 62: I'm not sure what this change to the comment is trying to say, but I feel like this comment is less helpful now

source/module.h
```diff
@@ -69,6 +69,14 @@ napi_status aws_napi_create_threadsafe_function(
     void *context,
     napi_threadsafe_function *result);
 
+/**
+ * Wrapper around napi_release_threadsafe_function,
+ * check the function before release it.
```
justinboswell, line 74: releasing

justinboswell (approved):
Fix-n-ship

source/module.h
TingDaoK: The previous comment is trying to make the threadsafe function a weak reference, and I removed the unref from the implementation. So, whenever we create a threadsafe function we need to release it.

source/module.h
```diff
@@ -69,6 +69,21 @@ napi_status aws_napi_create_threadsafe_function(
...
+ * check the function before releasing it.
+ */
+napi_status aws_napi_release_threadsafe_function(
+    napi_threadsafe_function function,
+    napi_threadsafe_function_release_mode mode);
+
+/**
+ * Wrapper around napi_unref_threadsafe_function,
+ * Incase release the threadsafe function from that function is needed, unref will let env go
+ * and the function will be cleaned up as env clean itself up
```
TingDaoK, line 83: Did an experiment that after unref, the `napi_finalize thread_finalize_cb` will be invoked even if no release ever invoked.🙆‍♂️

builder.json
```diff
@@ -90,6 +90,17 @@
...
             "run",
 	    "test:native"
+        ],
+        [
+            "npm",
+            "install",
+            "{project_dir}/samples/node/elasticurl/"
+        ],
+        [
+            "{python}",
```
justinboswell, line 100: YUSSSSSSS

source/http_stream.c
```diff
@@ -23,6 +23,9 @@ struct http_stream_binding {
     napi_threadsafe_function on_body;
     struct aws_http_message *response; /* used to buffer response headers/status code */
     struct aws_http_message *request;
+
+    uint64_t pending_length; /* used to check all the body callbacks for nodejs has been invoked */
```
justinboswell, line 27: "used to ensure that all of the body callbacks to node have been invoked"

source/http_stream.c
```diff
@@ -113,25 +116,34 @@ static int s_on_response_header_block_done(
...
     struct http_stream_binding *binding = context;
     struct on_body_args *args = user_data;
 
-    if (env) {
-        napi_value params[1];
-        const size_t num_params = AWS_ARRAY_SIZE(params);
-
-        AWS_NAPI_ENSURE(
-            env, napi_create_external_arraybuffer(env, args->chunk.buffer, args->chunk.len, NULL, NULL, &params[0]));
+    /* The callback has been invoked for the nodejs, update the pending length */
```
justinboswell, line 135: s/the//

source/http_stream.c
justinboswell: Also, looking at this, this value gets written to from both the node thread and the I/O threads. This should be an atomic.

source/http_stream.c
```diff
@@ -163,16 +179,26 @@ static void s_on_complete_call(napi_env env, napi_value on_complete, void *conte
...
 
-    if (env) {
-        napi_value params[1];
-        const size_t num_params = AWS_ARRAY_SIZE(params);
-
-        AWS_NAPI_ENSURE(env, napi_create_int32(env, args->error_code, &params[0]));
-        AWS_NAPI_ENSURE(
-            env,
-            aws_napi_dispatch_threadsafe_function(env, binding->on_complete, NULL, on_complete, num_params, params));
+    if (binding->pending_length) {
```
justinboswell, line 182: Do we still need this machinery if we have proven that the queued TS functions are in fact in order?

source/http_stream.c
```diff
@@ -163,16 +179,26 @@ static void s_on_complete_call(napi_env env, napi_value on_complete, void *conte
...
     }
 
+    napi_value params[1];
+    const size_t num_params = AWS_ARRAY_SIZE(params);
+
+    AWS_NAPI_ENSURE(env, napi_create_int32(env, args->error_code, &params[0]));
+    AWS_NAPI_ENSURE(
+        env, aws_napi_dispatch_threadsafe_function(env, binding->on_complete, NULL, on_complete, num_params, params));
+
+    /* No callbacks should happen now, cleanup all the threadsafe function */
```
justinboswell, line 198: functions

source/module.c
```diff
@@ -477,6 +489,7 @@ static void s_install_crash_handler(void) {
 static void s_napi_context_finalize(napi_env env, void *user_data, void *finalize_hint) {
     (void)env;
     (void)finalize_hint;
+    aws_event_loop_group_release(s_node_uv_elg);
```
justinboswell, line 492: 🤦

source/module.h
justinboswell: Note: If you want to release a thread safe function from within that thread safe function's callback, call unref instead, and the function will be finalized later by the environment.

source/http_stream.c
TingDaoK: What I found is that the complete callback **did** happen before some of body callbacks, even if from the C land, the complete will be the last one to be invoked. But, all the body callbacks happen in the right order. 
I still don't know about how nodejs dealing with it. 
But some guessing here: 
From their [samples](https://nodejs.github.io/node-addon-examples/special-topics/thread-safe-functions/ ), the TS functions are invoked as the same function in nodejs, which ensures the order. But, the different functions may be out of order.

justinboswell (approved):

PR #179, opened 2021-02-20
Add ArrayBuffer to supported MQTT payload types.

graebm:
This fixes aws-iot-device-sdk-js-v2, which shouldn't have been using the `mqtt.Payload` type anyway. But the type-checks used to pass, and now they will again.

Anyway, it's cool to support ArrayBuffer as an outgoing data type. We already supported it under the hood.

### DIFF
--- a/lib/browser/mqtt.ts
+++ b/lib/browser/mqtt.ts
@@ -141,7 +141,7 @@ class TopicTrie extends Trie<OnMessageCallback | undefined> {
  * @param payload The payload to convert
  * @internal
  */
-function normalize_payload(payload: Payload) : Buffer | string {
+function normalize_payload(payload: Payload): Buffer | string {
     if (payload instanceof Buffer) {
         // pass Buffer through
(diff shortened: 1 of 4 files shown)


lib/browser/mqtt.ts
```diff
@@ -395,8 +399,8 @@ export class MqttClientConnection extends BufferedEventEmitter {
                 }
                 resolve({
                     packet_id: packet
-                    ? (packet as mqtt.IUnsubackPacket).messageId
-                    : undefined,
+                        ? (packet as mqtt.IUnsubackPacket).messageId
+                        : undefined,
```
DavidOgunsAWS, line 403: Looks like indentation messed up here

DavidOgunsAWS (approved):

lib/browser/mqtt.ts
graebm: autoformatter did that 🤷‍♂️

PR #190, opened 2021-03-18
Timeout

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/native/aws_iot.ts
+++ b/lib/native/aws_iot.ts
@@ -198,6 +198,16 @@ export class AwsIotMqttConnectionConfigBuilder {
     }
 
     /**
+     * Configures the TCP socket timeout (in milliseconds)
+     * @param timeout_ms TCP socket timeout
+     * @deprecated
+     */
+    with_timeout_ms(timeout_ms: number) {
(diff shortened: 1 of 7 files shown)


lib/native/aws_iot.ts
```diff
@@ -198,11 +198,20 @@ export class AwsIotMqttConnectionConfigBuilder {
...
 
     /**
-     * Configures the TCP socket timeout (in milliseconds)
-     * @param timeout_ms TCP socket timeout
+     * Configures the PINGREQ response timeout (in milliseconds)
+     * @param ping_timeout PINGREQ response timeout
      */
-    with_timeout_ms(timeout_ms: number) {
-        this.params.timeout = timeout_ms;
+    with_ping_timeout_ms(ping_timeout: number) {
```
TingDaoK, line 204: Maybe we need some backward compatible way to deprecate the previous `with_timeout_ms` method?

justinboswell: Yeah, leave a method in here with the old signature, and call through to this one, and mark it `@deprecated`

justinboswell (approved):
Fix-n-shipperooni

PR #215, opened 2021-05-25
Exposing TlsContextOptions for websockets when using AwsIotMqttConnectionConfigBuilder

rccarper:
*Description of changes:*
* Exposing TLS Context Options for builder.
* npm audit fix


bretambrose (approved):
Let's switch the default on verify_peer in TlsCtxOptions to false as well and note it as a backwards incompatible change in the releases notes for both the crt and the sdk.

PR #216, opened 2021-05-25
stop using deprecated aws_off_t

graebm:
related:
https://github.com/awslabs/aws-c-common/pull/808
https://github.com/awslabs/aws-c-io/pull/386

### DIFF
--- a/source/io.c
+++ b/source/io.c
@@ -619,39 +619,35 @@ struct aws_napi_input_stream_impl {
     bool eos;          /* end of stream */
 };
 
-static int s_input_stream_seek(struct aws_input_stream *stream, aws_off_t offset, enum aws_stream_seek_basis basis) {
+static int s_input_stream_seek(struct aws_input_stream *stream, int64_t offset, enum aws_stream_seek_basis basis) {
     struct aws_napi_input_stream_impl *impl = stream->impl;
 
     int result = AWS_OP_SUCCESS;
     uint64_t final_offset = 0;
-    int64_t checked_offset = offset;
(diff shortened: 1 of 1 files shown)


source/io.c
```diff
@@ -619,39 +619,35 @@ struct aws_napi_input_stream_impl {
...
-            /* Offset must be negative, and must not be trying to go further back than the
-             * current length of the buffer, because those bytes have been purged
-             */
-            if (checked_offset > 0 || checked_offset == INT64_MIN || (uint64_t)(-checked_offset) > impl->buffer.len) {
-                result = AWS_IO_STREAM_INVALID_SEEK_POSITION;
-                goto failed;
-            }
-            final_offset = (uint64_t)impl->buffer.len - (uint64_t)(-checked_offset);
-            break;
+    if (basis == AWS_SSB_BEGIN) {
```
bretambrose, line 631: Should we make this part a public utility function in aws-c-*?

graebm: So I did that... then I went to incorporate it here, and it turns out that this function is doing something very different. While it clearly started as a copy/paste of that other function, it's doing something different now (can't seek backwards, after a read() it memmoves the remainder of the buffer to the start of the buffer, uses "offsets" to mean "offset from current read position" instead of "offset from start")

Anyway, I went and reverted the utility function I added to aws-c-io. Keeping this as is.

It was a great idea though! I totally agreed until I got to the 95% mark and was like waaaaait a minute.

bretambrose (approved):

PR #217, opened 2021-05-28
Fix message topic bug when using wildcard subscriptions

bretambrose:
* Use topic from C callback rather than original subscription value when forwarding a message


TingDaoK (approved):
LGTM

PR #219, opened 2021-06-02
Sigv4a

bretambrose:
* Sigv4a implementation


.github/workflows/ci.yml
```diff
@@ -3,18 +3,19 @@ name: CI
...
-      - '*'
-      - '!main'
+      - "*"
+      - "!main"
 
 env:
   BUILDER_VERSION: v0.8.11
   BUILDER_SOURCE: releases
   BUILDER_HOST: https://d19elf31gohf1l.cloudfront.net
-  PACKAGE_NAME: aws-crt-nodejs
```
TingDaoK, line 13: hmmm, I think I have the branch that only has the sigv4a stuff in it and those CI related stuff was removed [here](https://github.com/awslabs/aws-crt-nodejs-staging/pull/5/files). So, I guess we can squash that PR and cherry-pick the squashed commit.

TingDaoK: Or just remove all those changes in ci.yml, they are not needed for public repo. Either way.

justinboswell (approved):

source/module.c
```diff
@@ -368,6 +368,8 @@ napi_status aws_napi_dispatch_threadsafe_function(
     });
     /* Must always decrement the ref count, or the function will be pinned */
     napi_status release_status = napi_release_threadsafe_function(tsfn, napi_tsfn_release);
+    /* main thread can exit now */
+    AWS_NAPI_ENSURE(env, napi_unref_threadsafe_function(env, tsfn));
```
TingDaoK, line 372: This is not needed. It came from some testing I had from staging repo that I tired to fix the crash about threadsafe function. The public repo already have a better fix for the crash. Seems like I forgot to remove it from staging repo.

TingDaoK (approved):
Thanks!

PR #241, opened 2021-08-30
Callback rework and leak fixes

bretambrose:
* First draft to address various leaks:
* * flexible (can be called for any partial construction state) destroy functions for all argument structures; arg structures always use these for cleanup now
* * Use instance methods with the resolve/reject closures already bound in order to avoid accidentally capturing additional things in the environment in locally-created callbacks destined for native invocation
* * properly release the threadsafe function because it was created with a single reference that wasn't getting released (fix memory leak)  unref is not the right api to use.
* * mqtt connection builder to cache the created tls context on build.  This means you can't change tls options between invocations of build() but it also means you don't make a per-connection tls context which is very inefficient.
* * always clean up tls context options in aws_napi_io_tls_ctx_new()

In addition to fixing a number of leaks, the per-connection memory cost has gone from about 800K to 100K (almost entirely due to the tls context caching and leak fix).

There are some hidden gotchas with the API that will likely need backwards incompatible changes in order to address, as a followup.  In particular, failing to connect on the initial call will turn the connection into something completely useless.  This is not intended.  Additionally, calling disconnect does the same thing - it completely wipes all of the callbacks from native to node (by calling close()) and there's no way to restore them outside of making a new connection.  Both of these are incorrect behavior that is out of sync with the other CRTs: it should be possible to succesfully call connect() on a disconnected connection as well as a connection that failed its initial connect() call.  

Additionally, we don't have a way of detecting when we can clean up the connection outside of an explicit disconnect.  Currently I would like to see close() moved from private to public to allow applications to guarantee everything is cleaned up properly no matter what the connection state.  Unfortunately, that's a backwards incompatible change and will need to wait for the future.

Finally, all the non-mqtt bindings need their own pass to apply the same patterns.


lib/native/aws_iot.ts
```diff
@@ -283,7 +283,16 @@ export class AwsIotMqttConnectionConfigBuilder {
             throw 'client_id and endpoint are required';
         }
 
-        this.params.tls_ctx = new io.ClientTlsContext(this.tls_ctx_options);
+        /*
+         * By caching and reusing the tls context we get an enormous memory savings on a per-connection basis.
+         * The tradeoff is that you can't modify tls options in between calls to build.
+         * Previously we were making a new one with every single connection which had a huge negative impact on large
+         * scale tests.
+         */
```
justinboswell, line 291: Can you add this to the tsdocs so customers get it?

lib/native/mqtt.ts
```diff
@@ -237,29 +245,20 @@ export class MqttClientConnection extends NativeResourceMixin(BufferedEventEmitt
...
-            }
-
             try {
                 crt_native.mqtt_client_connection_connect(
                     this.native_handle(),
                     this.config.client_id,
                     this.config.host_name,
                     this.config.port,
+                    resolve,
+                    reject,
```
justinboswell, line 255: After thinking about this for 2 days, what if we used [bind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) to create the single closure that needs to be passed to native code here. I really, really don't like the extra ref counting in native code, especially when all of the references have the same lifetime.
e.g.
```typescript
this._on_connect_callback.bind(this, resolve, reject)
```

justinboswell (changes requested):

lib/native/mqtt.ts
bretambrose: I'll give it a try and see how it goes.

justinboswell (approved):

PR #247, opened 2021-09-07
Use OpenSSL baked into node instead of AWS LC

justinboswell:
Current outstanding issue: If the openssl version does not match the API version in node (1.1.x), then you will get crashes because libcrypto will be loaded from the system. Most likely, the fix will involve mandating openssl 1.1 at the aws-c-cal level. If it can't be found, then aws-lc must be used.


bretambrose:
Closing for now

PR #250, opened 2021-09-08
Split debug info out from node module

justinboswell:
Depends on:
* https://github.com/awslabs/aws-crt-nodejs/pull/248


TingDaoK (approved):

bretambrose (commented):
When a native crash happens after this change, will a function-annotated call stack be visible to a user consuming it in a standard way (not building it manually)?

justinboswell:
> When a native crash happens after this change, will a function-annotated call stack be visible to a user consuming it in a standard way (not building it manually)?

If they have installed the symbol package, yes.

PR #252, opened 2021-09-09
To solve the npm package size issue

TingDaoK:
*Issue #, if available:*
- The current aws-crt npm package size is too huge(200 MB)

*Description of changes:*
- Remove C source code from npm package
- Publish C source code to S3 with publish pipeline, and the build script will download it when needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ You can either add it to package.json (if using a tool like webpack), or just im
 npm install aws-crt
 ```
 
-To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_SOURCE_CODE_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
+To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_BINARY_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
 
 ### Debug C part of code
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ You can either add it to package.json (if using a tool like webpack), or just im
 npm install aws-crt
 ```
 
-To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_BINARY_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
+To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, the install script will pull the source from S3 bucket. In case of no public internet access, you can specify the "CRT_BINARY_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
 
 ### Debug C part of code
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ You can either add it to package.json (if using a tool like webpack), or just im
 npm install aws-crt
 ```
 
+To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_SOURCE_CODE_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz, the sha256 checksum is https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256
+
 ### Debug C part of code
 
 After building the package locally, use ```node ./scripts/build.js --debug``` to enable debug. Then, attach any C debugger to use node to run `jest`
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ You can either add it to package.json (if using a tool like webpack), or just im
 npm install aws-crt
 ```
 
-To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_SOURCE_CODE_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz, the sha256 checksum is https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256
+To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_SOURCE_CODE_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
 
 ### Debug C part of code
 
--- a/continuous-delivery/pack.sh
+++ b/continuous-delivery/pack.sh
@@ -7,6 +7,11 @@ git describe --tags
 CURRENT_TAG=$(git describe --tags | cut -f2 -dv)
 # convert v0.2.12-2-g50254a9 to 0.2.12
 CURRENT_TAG_VERSION=$(git describe --tags | cut -f1 -d'-' | cut -f2 -dv)
+# if there's a hash on the tag, then this is not a release tagged commit
+if [ "$CURRENT_TAG" != "$CURRENT_TAG_VERSION" ]; then
+    echo "Current tag version is not a release tag, cut a new release if you want to publish."
+    exit 1
+fi
 
 # go to previous directory
 cd ..
--- a/continuous-delivery/pack.sh
+++ b/continuous-delivery/pack.sh
@@ -12,16 +12,25 @@ CURRENT_TAG_VERSION=$(git describe --tags | cut -f1 -d'-' | cut -f2 -dv)
 cd ..
 # native source code
 tar -cvzf aws-crt-$CURRENT_TAG_VERSION-source.tgz aws-crt-nodejs/crt
+# sha1 checksum
+SOURCE_SHA1=$(sha1sum aws-crt-$CURRENT_TAG_VERSION-source.tgz | awk '{print $1}')
+echo $SOURCE_SHA1 > aws-crt-$CURRENT_TAG_VERSION-source.sha1
 # omnibus package
 tar -cvzf aws-crt-$CURRENT_TAG_VERSION-all.tgz aws-crt-nodejs/
+# sha1 checksum
+SOURCE_SHA1=$(sha1sum aws-crt-$CURRENT_TAG_VERSION-all.tgz | awk '{print $1}')
+echo $SOURCE_SHA1 > aws-crt-$CURRENT_TAG_VERSION-all.sha1
 # binaries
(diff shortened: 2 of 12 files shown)


continuous-delivery/pack.sh
```diff
@@ -0,0 +1,39 @@
...
+# UNZIP="unzip_pack"
+# mkdir $UNZIP
+# tar -xf aws-crt-$CURRENT_TAG_VERSION.tgz -C $UNZIP
+# PACK_FILE_SIZE_KB=$(du -sk $UNZIP | awk '{print $1}')
+# if expr $PACK_FILE_SIZE_KB \< 3000000 ; then
+#     # True, the package size is smaller than 3 MB, return 0
+#     exit 0
+# fi
+# # False
+# echo "Package size is too large"
```
TingDaoK, line 38: for now, the unzip size is around 10.9 includes all the binaries. And linux binaries are larger than expected. Comment this check out, until Monday, before merge.

scripts/build.js
```diff
@@ -5,28 +5,132 @@
...
+
+async function fetch_native_code(url, version, path) {
+    const source_URL = url + "aws-crt-" + version + "-source.tgz"
+    const tarball_path = path + "source.tgz"
+    return new Promise((resolve, reject) => {
+        download_file(source_URL, tarball_path).then(() => {
+            // download checksum
+            const source_checksum_URL = url + "aws-crt-" + version + "-source.sha1"
+            check_checksum(source_checksum_URL, tarball_path)
+            // check_checksum(source_checksum_URL, "/Users/dengket/Downloads/aws-crt-1.9.2-source.tgz")
```
TingDaoK, line 69: something is wrong! The cloudfront link works perfectly on my browser, downloaded the latest .tgz file. However, the nodejs axios gives me a cached or old version of the file? I checked the returned header, it apparently not the latest one. I don't really know what happens.

scripts/build.js
```diff
@@ -5,28 +5,132 @@
...
-});
-buildSystem.build();
+if (!fs.existsSync("crt/")) {
+    const tmp_path = path.join(__dirname, uuidv4() + "temp/");
+    fs.mkdirSync(tmp_path);
+
+    // There is no native code, we are not building from source.
+    (async () => {
+        // AWS common runtime aws-crt-nodejs cloudfront distribution.
+        const url = "http://d332vdhbectycy.cloudfront.net/";
```
TingDaoK, line 121: We can use github release asset instead of cloudfront deployment when we release. But, AWS is easier to test for now.

scripts/build.js
TingDaoK: I just tested today, it works now🙃

package.json
```diff
@@ -47,6 +47,9 @@
     "crypto-js": "^4.0.0",
     "fastestsmallesttextencoderdecoder": "^1.0.22",
     "mqtt": "^4.2.8",
+    "tar": "6.1.11",
```
bretambrose, line 50: should these be ^

bretambrose (commented):
Seems reasonable but unless I'm missing something we should test this cloudfront-sourced build on all three platforms just as a sanity check.

TingDaoK:
> Seems reasonable but unless I'm missing something we should test this cloudfront-sourced build on all three platforms just as a sanity check.

It's a bit tricky to test it. What I did is download the packed tarball from `npm pack`, and test it by:
- Remove the binary
- Run `node scripts/install.js` locally.
I tested it on mac manually, and it works. I can test it on linux and windows for sure.

scripts/build.js
```diff
@@ -5,28 +5,135 @@
...
+}
+
+async function fetch_native_code(url, version, path) {
+    const source_URL = url + "aws-crt-" + version + "-source.tgz"
+    const tarball_path = path + "source.tgz"
+    return new Promise((resolve, reject) => {
+        download_file(source_URL, tarball_path).then(() => {
+            // download checksum
+            const source_checksum_URL = url + "aws-crt-" + version + "-source.sha1"
+            check_checksum(source_checksum_URL, tarball_path)
```
graebm, line 68: seems more secure to embed the checksum in the NPM package. That way an "adversary" can't switch out both files on you using their internet magic

also sha-512 probably more secure. I hear people don't like sha-1 anymore

I am not a security professional

TingDaoK: I tested on windows, and got a mismatch of the checksum once. I retried to run the script again, it works. Probably need to apply some retry strategy? Something like simple retries three times(re-download the source tarball and re-calculate the checksum).

scripts/build.js
TingDaoK: I think we can ignore the caching problem, as when we publish for real, we will have different version number every time. It will be different urls, and get around the caching issue.

scripts/build.js
```diff
@@ -5,28 +5,140 @@
...
+    const tmp_path = path.join(__dirname, uuidv4() + "temp/");
+    fs.mkdirSync(tmp_path);
+
+    // There is no native code, we are not building from source.
+    (async () => {
+        // AWS common runtime aws-crt-nodejs cloudfront distribution.
+        let url = "https://d332vdhbectycy.cloudfront.net/";
+        if (process.env.CRT_SOURCE_CODE_HOST) {
+            // Use the host specified by user
+            url = process.env.CRT_SOURCE_CODE_HOST;
```
graebm, line 129: couple things:
- maybe call this `host` instead of `url`, since the env-var to override is is called "host"?
- don't have the trailing slash be part of this. trailing slash is part of the path, not the host

scripts/build.js
```diff
@@ -5,28 +5,140 @@
...
+        }
+        let rawdata = fs.readFileSync('package.json');
+        let package = JSON.parse(rawdata);
+        const version = package["version"];
+        fetch_native_code(url, version, tmp_path).then(() => {
+            // Clean up temp directory
+            fs.rmSync(tmp_path, { recursive: true });
```
graebm, line 136: to be improved in the future: probably better to cache the download, but then we'd need integrity checks to ensure the cached version is good, so this works for now. Here's [prebuild-install](https://github.com/prebuild/prebuild-install#cache) explaining their caching strategy

package.json
```diff
@@ -47,6 +47,9 @@
     "crypto-js": "^4.0.0",
     "fastestsmallesttextencoderdecoder": "^1.0.22",
     "mqtt": "^4.2.8",
+    "tar": "^6.1.11",
+    "fs-extra": "^10.0.0",
+    "checksum": "^1.0.0",
```
graebm, line 52: these are dependencies now for browser too, right? any way around that?

graebm: It would be nice to dump these new dependencies in a future pass:

fs-extra: looks like it's being used for recursive copy. we could implement our own. looks like Node 16 added a more useful [fs.cpSync()](https://nodejs.org/api/fs.html#fs_fs_cpsync_src_dest_options), but we're currently supporting Node 10

checksum: don't think we need this? Here's an example using node's built in [crypto](https://nodejs.org/api/crypto.html#crypto_crypto_createhash_algorithm_options) module

scripts/build.js
```diff
@@ -5,28 +5,140 @@
...
+        });
+    });
 }
 
-// Convert any -D arguments to this script to cmake -D arguments
-for (const arg of process.argv) {
-    if (arg.startsWith('-D')) {
-        const option = arg.substring(2).split('=')
-        options[option[0]] = option[1]
+async function check_checksum(url, loacl_file) {
```
graebm, line 39: trivial: ~loacl_file~ -> local_file

package.json
TingDaoK: Yeah, I found that in Node 16 too. I'll try to implement it. But, I don't think we can avoid `tar`...

README.md
```diff
@@ -40,6 +40,8 @@ You can either add it to package.json (if using a tool like webpack), or just im
 npm install aws-crt
 ```
 
+To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_SOURCE_CODE_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
```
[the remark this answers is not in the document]
TingDaoK: lol, I was confused by the "binary" as those ".node" files are things we talked about a lot.

README.md
```diff
@@ -40,6 +40,8 @@ You can either add it to package.json (if using a tool like webpack), or just im
 npm install aws-crt
 ```
 
+To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, npm will pull the source from S3 bucket. In case, you don't have the public internet, you can specify the "CRT_BINARY_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
```
AllanZhengYP, line 43:
```suggestion
To reduce the size of package, we put the C source code in the S3 bucket. If the platform you are using doesn't have the prebuilt binary, the install script will pull the source from S3 bucket. In case of no public internet access, you can specify the "CRT_BINARY_HOST" environment variable for the host of the source code. The build script will fetch source code from that host instead. To fetch the source from S3, you can reach to the cloudfront distribution (Only works for version after v1.9.2) `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.tgz`, the sha256 checksum `https://d332vdhbectycy.cloudfront.net/aws-crt-<version>-source.sha256`
```

continuous-delivery/pack.sh
```diff
@@ -0,0 +1,50 @@
...
+npm pack --unsafe-perm
+cp aws-crt-*.tgz ..
+
+# Check unzip npm package size
+cd ..
+UNZIP="unzip_pack"
+mkdir $UNZIP
+tar -xf aws-crt-$CURRENT_TAG_VERSION.tgz -C $UNZIP
+PACK_FILE_SIZE_KB=$(du -sk $UNZIP | awk '{print $1}')
+if expr $PACK_FILE_SIZE_KB \> 12000 ; then
```
AllanZhengYP, line 45: Very nice to have this check!

continuous-delivery/pack.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+set -ex
+
+# force a failure if there's no tag
+git describe --tags
+# now get the tag
+CURRENT_TAG=$(git describe --tags | cut -f2 -dv)
+# convert v0.2.12-2-g50254a9 to 0.2.12
+CURRENT_TAG_VERSION=$(git describe --tags | cut -f1 -d'-' | cut -f2 -dv)
```
AllanZhengYP, line 9: May be check if the tagged commit at current head? This can prevent updating binaries without bumping versions.

scripts/build.js
graebm: after talking with wizards, they reminded me that `https` prevents a evil wizards from spoofing the host

package.json
```diff
@@ -47,6 +47,7 @@
     "crypto-js": "^4.0.0",
     "fastestsmallesttextencoderdecoder": "^1.0.22",
     "mqtt": "^4.2.8",
+    "tar": "^6.1.11",
```
AllanZhengYP, line 50: We saw customer complaining of the deprecated dependencies too, are those tackled already? 
https://github.com/aws/aws-sdk-js-v3/issues/2750

scripts/build.js
```diff
@@ -3,30 +3,144 @@
...
+}
 
-let options = {
-    CMAKE_EXPORT_COMPILE_COMMANDS: true,
-    CMAKE_JS_PLATFORM: os.platform,
-    CMAKE_JS_ARCH: os.arch,
-    BUILD_TESTING: 'OFF',
-    CMAKE_INSTALL_PREFIX: 'crt/install',
-    CMAKE_PREFIX_PATH: 'crt/install',
+async function checkChecksum(url, local_file) {
```
AllanZhengYP, line 39: I don't think you need `async` here

scripts/build.js
```diff
@@ -3,30 +3,144 @@
...
+const fs = require("fs");
+const crypto = require('crypto');
 const process = require("process");
+const path = require("path");
+
 const cmake = require("cmake-js");
+const axios = require("axios");
+const tar = require('tar');
+
+async function downloadFile(fileUrl, outputLocationPath) {
```
AllanZhengYP, line 15: I don't think you need `async` here

scripts/build.js
```diff
@@ -3,30 +3,144 @@
...
+            const sourceChecksumURL = `${url}/aws-crt-${version}-source.sha256`
+            checkChecksum(sourceChecksumURL, tarballPath)
+            tar.x({ file: tarballPath, strip: 2, C: nativeSourceDir })
+                .then(() => { resolve(); })
+                .catch((err) => { reject(err); })
+        }).catch((err) => {
+            reject(err)
+        })
+    });
+}
```
AllanZhengYP, line 83: You async-await instead:
```suggestion
async function fetchNativeCode(url, version, path) {
    const sourceURL = `${url}/aws-crt-${version}-source.tgz`
    const tarballPath = path + "source.tgz";
    await downloadFile(sourceURL, tarballPath);
    const sourceChecksumURL = `${url}/aws-crt-${version}-source.sha256`;
    await checkChecksum(sourceChecksumURL, tarballPath);
    await tar.x({ file: tarballPath, strip: 2, C: nativeSourceDir });
}
```

scripts/build.js
```diff
@@ -3,30 +3,144 @@
...
+process.chdir(workDir);
+const nativeSourceDir = "crt/"
+
+if (!fs.existsSync(nativeSourceDir)) {
+    const tmpPath = path.join(__dirname, `temp${crypto.randomBytes(16).toString("hex")}/`);
+    fs.mkdirSync(tmpPath);
+    fs.mkdirSync(nativeSourceDir);
+
+    // There is no native code, we are not building from source.
+    (async () => {
```
AllanZhengYP, line 123: This IIFE looks weird, why not bundle all the code in an async function?

```javascript
async function run() {
    // All the top-level code
}

run();
```

AllanZhengYP (commented):
Looks good generally, left a few comments

scripts/build.js
```diff
@@ -3,30 +3,144 @@
...
-    debug: process.argv.includes('--debug'),
-    cMakeOptions: options,
-});
-buildSystem.build();
+// Makes sure the work directory is what we need
+const workDir = path.join(__dirname, "../")
+process.chdir(workDir);
+const nativeSourceDir = "crt/"
+
+if (!fs.existsSync(nativeSourceDir)) {
```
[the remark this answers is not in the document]
TingDaoK: I removed it after local build...

continuous-delivery/pack.sh
TingDaoK: I removed a check that was there for testing. Added it back to check if the tag is release. So, in that case, we can prevent updating binaries without proper release, which will also prevent the version is not updated, I think.

package.json
TingDaoK: I saw those warnings as well, don't know how to solve them... `npm update` and `npm audit fix` is not helping... But, we can probably tackle them after

graebm (approved):

PR #255, opened 2021-09-16
fix code deployment

TingDaoK:
- We have more stuff from previous steps (aws-crt-\<version\>-source.tgz, aws-crt-\<version\>-binary.tgz). The previous wildcard is not working anymore, need the package ends up with numbers (aws-crt-\<version\>.tgz)


bretambrose (commented):
What's the exact pattern you're trying to match, this doesn't quite look right.  It looks like "everything that starts with 'aws-crt' and ends with a single digit.  It might work but in terms of intent and readability wouldn't it be better to match the exact file pattern?

TingDaoK:
> What's the exact pattern you're trying to match, this doesn't quite look right. It looks like "everything that starts with 'aws-crt' and ends with a single digit. It might work but in terms of intent and readability wouldn't it be better to match the exact file pattern?

I want to match the file `aws-crt-\<version\>.tgz`, and the version can be digitals and `.` and it's not sure about how many digital it will have. I don't know what's the best way to match that. This change will rule out the other files ends with `-source.tgz` or `-binary.tgz` we included from the [previous steps](https://github.com/awslabs/aws-crt-nodejs/blob/main/continuous-delivery/pack.sh#L20-L32).

sbSteveK (approved):

PR #260, opened 2021-09-20
rmdir recursive is after node 12.10

TingDaoK:
*Issue #, if available:*

- Failed to build on Raspberry PI with node 10, as rmdir recursive is introduced at node 12.10. And it's not documented from node as it's deprecated now. :(

*Description of changes:*

- Implement our own recursive rmdir


scripts/build.js
```diff
@@ -12,6 +12,27 @@ const cmake = require("cmake-js");
 const axios = require("axios");
 const tar = require('tar');
 
+function rmRecursive(path) {
+    var files = [];
+    if (fs.existsSync(path)) {
+        if (fs.lstatSync(path).isDirectory()) {
+            files = fs.readdirSync(path);
```
justinboswell, line 19: Do we need to filter out `.` and `..` so we don't wipe a drive/mount point?

TingDaoK: I feel it's more like a user's responsibility to not delete the thing they need. Like this function is just a tool, user should be responsible to not use it wipe mount point.

TingDaoK: changed mind, as yeah, rm doesn't rm . and .. as well

justinboswell (approved):

PR #271, opened 2021-11-04
Size Check Script added

sbSteveK:
Testing size check for CI


TingDaoK:
`.builder/actions/__pycache__/crt_size_check.cpython-39.pyc` this seems like should not be included?

bretambrose (commented):
Should this be checking the overall dist size as well?

graebm:
> `.builder/actions/__pycache__/crt_size_check.cpython-39.pyc` this seems like should not be included?

Yeah, you'll want to put `*.pyc` in your [global git ignore](https://stackoverflow.com/a/7335487)

bretambrose (approved):

TingDaoK:
> Should this be checking the overall dist size as well?

Maybe not? We only include the compiled binaries in the npm package, and that's what the check for. It's not meaningful to check the size of `lib/` and `include/`

PR #273, opened 2021-11-09
Changed crt size check to only include NPM packaged files

sbSteveK:


.builder/actions/crt_size_check.py
```diff
@@ -9,33 +9,29 @@
...
+        for root, dirs, files in os.walk(os.path.join(env.project.path, 'dist/native/')):
 
             for f in files:
                 fp = os.path.join(root, f)
-                current_folder_size += os.path.getsize(fp)
+                total_size += os.path.getsize(fp)
+
+        print(f"/native files total size: {total_size} bytes")
 
+        for root, dirs, files in os.walk(os.path.join(env.project.path, 'dist/')):
```
TingDaoK, line 24: you don't need a separate for loop here, the `.node` file will always be in `dist/`

.builder/actions/crt_size_check.py
```diff
@@ -9,33 +9,29 @@
...
         # Maximum dist folder size in bytes
-        max_dist_size = 50_000_000
-        # Size of current folder
-        current_folder_size = 0
-        # Total size of files in dist folder
         total_size = 0
 
-        for root, dirs, files in os.walk(os.path.join(env.project.path, 'dist/')):
-            current_folder_size = 0
+        for root, dirs, files in os.walk(os.path.join(env.project.path, 'dist/native/')):
```
TingDaoK, line 16: I was wrong, maybe just leave `lib` and `include` out, which are C code and libraries that will not be included in the package we are shipping to npm

.builder/actions/crt_size_check.py
```diff
@@ -9,33 +9,29 @@
...
                 fp = os.path.join(root, f)
-                current_folder_size += os.path.getsize(fp)
+                total_size += os.path.getsize(fp)
+
+        print(f"/native files total size: {total_size} bytes")
 
+        for root, dirs, files in os.walk(os.path.join(env.project.path, 'dist/')):
+            for f in files:
                 if 'aws-crt-nodejs.node' == f:
+                    fp = os.path.join(root, f)
```
justinboswell, line 27: what about the .js files in here? Those are also distributed

sbSteveK: Just spoke with Dengke. Going to change it to include all files in the dist folder except the ones in lib and include.

justinboswell: actually, come to think of it, everything in dist/ is shipped

justinboswell (approved):

PR #276, opened 2021-11-22
Update cmake-js to latest version

orgads:
This removes transient dependency on deprecated request package.


bretambrose:
We cannot take 6.2.1; it is broken on old MSVC installations.  See this issue here (https://github.com/cmake-js/cmake-js/issues/249) as well as the request to make a new release here: https://github.com/cmake-js/cmake-js/issues/265.  If 6.2.2 is released and passes our CI, we will update to it.

PR #278, opened 2021-11-22
Upgrade aws-sdk to v3 + dependencies in browser samples

orgads:


bretambrose (commented):
To be honest, it's not clear to me where these PRs are going or the need to do them.

I don't know Javascript, it's ecosystem, or its development culture, but it feels fundamentally unhealthy to perpetually get caught up in the "major version treadmill" that the JS ecosystem seems to experience.  Ideally, we'd like to limit version upgrades to  security issues.

bretambrose:
Closing absent a scope-down to just the v3 aws sdk bump (which is a nice upgrade, to be fair).

PR #284, opened 2021-12-20
Fix unresolved promise on disconnect

richarddd:
*Description of changes:*
When calling: `await client.disconnect()` the promise never resolves.

This is due to how the API is a bit misleading on how callbacks are supposed to be passed to the mqtt client:
https://github.com/mqttjs/MQTT.js/blob/master/lib/client.js#L978-L984

This PR fixes the disconnect promise callback

### DIFF
--- a/lib/browser/mqtt.ts
+++ b/lib/browser/mqtt.ts
@@ -486,7 +486,7 @@ export class MqttClientConnection extends BufferedEventEmitter {
     */
     async disconnect() {
         return new Promise((resolve) => {
-            this.connection.end(null, resolve)
+            this.connection.end(undefined, resolve)
         });
     }
 }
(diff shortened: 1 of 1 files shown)


bretambrose:
This approach fails typescript compilation so we can't merge as is.

I ran into this a while back and added a potential fix in a PR that later got shelved due to risk/relevance to the original issue: https://github.com/awslabs/aws-crt-nodejs/pull/254/files#diff-815e775af9421298456e403cb484069f9d9c90e06b141d05aa34ad79874dad67R387

Can you try that version of the fix and see if it does what you need?  The .end() implementation is absurdly complex (possibly trying to maintain backwards compatibility with previous signatures) and very difficult to mentally "simulate."  

Semantically, our browser disconnect() needs to match our nodejs disconnect() (which is a complete teardown and state erase).  I'm worried from your use case description that a complete tear-down would not work for you.

bretambrose:
Merged in https://github.com/awslabs/aws-crt-nodejs/pull/285

PR #295, opened 2022-02-09
Support PKCS#11 for mutual TLS on Unix platforms

graebm:
Add bindings for new PKCS#11 functionality.

Related: https://github.com/awslabs/aws-c-io/pull/451


source/io.c
```diff
@@ -342,24 +466,100 @@ napi_value aws_napi_io_tls_ctx_new(napi_env env, napi_callback_info info) {
...
+                    goto cleanup;
+                }
+                pkcs11_options.slot_id = &pkcs11_slot_id;
+            }
+        }
+
+        /* clang-format off */
+
+        #define PARSE_PKCS11_OPTION_STR(property_name, option_cursor, storage_buffer)                                  \
+        do {                                                                                                           \
```
bretambrose, line 543: Can't we just use a scoped block here?

source/logger.c
```diff
@@ -213,29 +207,47 @@ static void s_threadsafe_log_call(napi_env env, napi_value node_log_fn, void *co
...
+        goto done;
+    }
+
     napi_value node_process = NULL;
-    AWS_NAPI_CALL(env, napi_get_named_property(env, node_global, "process", &node_process), { goto done; });
+    if (napi_ok != napi_get_named_property(env, node_global, "process", &node_process)) {
+        goto done;
+    }
+
+    for (struct aws_linked_list_node *list_node = aws_linked_list_rbegin(&msgs);
```
bretambrose, line 223: I'd add a comment note for why reverse iteration is correct

bretambrose (approved):

source/io.c
graebm: it's the classic "put your macro in a `do{}while(0)` so users are forced to put semicolon after it" trick

but yeah, confusing, and this macro not a global helper

I'll simplify

source/logger.c
graebm: 😱 OH NO THIS IS NOT CORRECT AT ALL
great catch. I have failed this interview

PR #305, opened 2022-03-15
bump up package size check

graebm:
A release just failed because we were 72 bytes over the cap.

Bump the cap up a smidge so we can release a security fix.

Long term, we should be doing these checks with each commit to main, not as something that fails at release time.


TingDaoK (approved):
lol

xomyakus (approved):

PR #316, opened 2022-05-05
Support for Apple M1

graebm:
**Issue:**
NPM package lacked arm64 binaries for macOS, so it didn't work as nicely on newer M1 Macs.

**Description of Changes:**
NPM package now includes arm64 binaries for macOS.

* Support cross-compiling on macOS, so any machine can make an x64 or arm64 binary.
  * Can't just use an arm64 machine because we only have access to x64 macOS in our release pipeline.
  * Decided against creating a single "universal" binary. It's a hassle to build (need to compile for both x64, then arm64, then glue them together with lipo) and would increase iteration time when developing on a Mac. And since our code [picks which binary to load](https://github.com/awslabs/aws-crt-nodejs/blob/12d30161dd8a3f78c0428caafe6ba1bd12b43288/lib/native/binding.js#L22), having distinct binaries seems simpler for now.
* A single `build-osx-x64-arm64.sh` job builds both x64 and arm64 binaries.
  * Decided against a separate arm64 job since we only have an x64 machine right now. So `npm install` isn't going to naturally spit out an arm64 build. Therefore we must cross-compile. And it seemed simplest to do that after the normal `npm install` that pulls dependencies but also creates the x64 binary.
* No more support for binaries in the `dist/bin/native/` folder 
  * This was leading to weird errors. The platform must be known at build time now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,9 +96,9 @@ aws_use_package(aws-c-auth REQUIRED)
 aws_use_package(aws-checksums REQUIRED)
 target_link_libraries(${PROJECT_NAME} PRIVATE ${CMAKE_JS_LIB} ${DEP_AWS_LIBS})
 
-if (CMAKE_JS_PLATFORM AND CMAKE_JS_ARCH)
+if (CMAKE_JS_PLATFORM AND NODE_ARCH)
     # When run from build.ts, these are set
-    set(destination bin/${CMAKE_JS_PLATFORM}-${CMAKE_JS_ARCH})
+    set(destination bin/${CMAKE_JS_PLATFORM}-${NODE_ARCH})
 else()
     # If not, default to "native"
     set(destination bin/native)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,8 +100,7 @@ if (CMAKE_JS_PLATFORM AND NODE_ARCH)
     # When run from build.ts, these are set
     set(destination bin/${CMAKE_JS_PLATFORM}-${NODE_ARCH})
 else()
-    # If not, default to "native"
-    set(destination bin/native)
+    message(FATAL_ERROR "CMAKE_JS_PLATFORM and NODE_ARCH must be set")
 endif()
 
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/aws-crt-nodejs.node"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,11 @@ cmake_minimum_required(VERSION 3.1)
 project(aws-crt-nodejs C)
 option(BUILD_DEPS "Builds aws common runtime dependencies as part of build, only do this if you don't want to control your dependency chain." ON)
 
+option(CMAKE_JS_PLATFORM "Target platform. Should match node's os.platform()")
+if (NOT CMAKE_JS_PLATFORM)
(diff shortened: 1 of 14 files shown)


scripts/build.js
```diff
@@ -99,32 +99,48 @@ async function fetchNativeCode(url, version, path) {
...
+            CMAKE_JS_PLATFORM: platform,
+            BUILD_TESTING: 'OFF',
+            CMAKE_INSTALL_PREFIX: 'crt/install',
+            CMAKE_PREFIX_PATH: 'crt/install',
+        }
+    }
+
+    if (platform === 'darwin') {
+        // Node calls it "x64" but Apple calls it "x86_64", they both agree on "arm64" though
+        options.cMakeOptions.CMAKE_OSX_ARCHITECTURES = (arch === 'x64') ? 'x86_64' : arch;
```
TingDaoK, line 128: Just curious, this add a CMAKE_OSX_ARCHITECTURES cmake option, but the `options.arch` is still the old arch. Is that what we want?

CMakeLists.txt
```diff
@@ -96,13 +101,7 @@ aws_use_package(aws-c-auth REQUIRED)
...
 target_link_libraries(${PROJECT_NAME} PRIVATE ${CMAKE_JS_LIB} ${DEP_AWS_LIBS})
 
-if (CMAKE_JS_PLATFORM AND CMAKE_JS_ARCH)
-    # When run from build.ts, these are set
-    set(destination bin/${CMAKE_JS_PLATFORM}-${CMAKE_JS_ARCH})
-else()
-    # If not, default to "native"
-    set(destination bin/native)
-endif()
+set(destination bin/${CMAKE_JS_PLATFORM}-${NODE_ARCH})
```
TingDaoK, line 104: Why we use `NODE_ARCH` instead?

scripts/build.js
TingDaoK: Is that because from the [doc](https://github.com/cmake-js/cmake-js#runtimes) from cmakejs? (I am not sure I am looking at the right place though) 
> Notice: on non-Windows systems the C++ toolset's architecture's gonna be used despite this setting. 

If so, can we add a comment about it?

CMakeLists.txt
graebm: NODE_ARCH is a variable that is always set by cmake-js

Previously, we'd been using a variable we invented named `CMAKE_JS_ARCH`. There didn't seem to be a need for inventing our own variable, so I stopped using it

scripts/build.js
graebm: Ha, I didn't even notice that comment. I just found out cmake-js wasn't passing the right flags and knew how to fix it.
I added a comment, along with that citation

CMakeLists.txt
TingDaoK: So, `NODE_ARCH` comes from the `options.arch`?

TingDaoK (approved):
LGTM

CMakeLists.txt
graebm: yeah, passing `options.arch` cmake-js leads to it setting this variable.

There are a few other variables from cmake-js that we're already using in our `CMakeLists.txt`, like: `CMAKE_JS_LIB` and `CMAKE_JS_INC`. I don't know whey they named this one `NODE_ARCH` instead of `CMAKE_JS_ARCH`, the naming is inconsistent. And I don't know why we chose to name a variable that we invented `CMAKE_JS_ARCH`, it makes it look like that variable is coming from cmake-js but it's not.

oh well. simpler now

PR #319, opened 2022-05-19
Update Auth Interface 

xiazhvera:
*Description of changes:*
- Update Auth.ts interface to share between node and browser api
- Implement credentials provider with static credentials provider, AWS Cognito Credentials provider, and cached credentials provider. The provider should handle the issue where the session expiration for web connections.


lib/browser/auth.ts
```diff
@@ -11,17 +11,37 @@
...
+    /** external credential provider */
+    aws_provider ?: any
+}
+
+/**
+ * Configuration for use in browser credential
+ *
+ * @category Auth
+ */
+export interface AwsSigningConfig extends AwsSigningConfigBase{
```
bretambrose, line 39: Why does this need to be exported?  It's not something we'd ever want external users to submit given its restriction to just websocket handshakes

lib/common/auth.ts
```diff
@@ -0,0 +1,142 @@
...
+    /** The region to sign against */
+    region: string;
+    /** Name of service to sign a request for */
+    service?: string;
+    /**
+     * Date and time to use during the signing process. If not provided then
+     * the current time in UTC is used. Naive dates (lacking timezone info)
+     * are assumed to be in local time
+     */
+    date?: Date;
```
bretambrose, line 104: doesn't this make the date addition to browser AwsSigningConfig redundant or  confusing design-wise.

sbSteveK (approved):

lib/common/auth.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @internal
+ */
+export type StringLike = string | ArrayBuffer | DataView;
+
+/**
+ * AWS signing algorithm enumeration.
+ *
+ * @category Auth
+ */
+export enum AwsSigningAlgorithm {
```
bretambrose, line 24: I'm not sure we want these to be shared.  We're never going to support these options in the browser so it seems weird to put them in the common definitions.

lib/common/auth.ts
```diff
@@ -0,0 +1,142 @@
...
+    create_signing_config ?: ()=> AwsSigningConfigBase;
+
+    /**
+     * (Optional) override for the service name used in signing the websocket handshake.  Leave null to use the
+     * default (iotdevicegateway)
+     */
+     service?: string;
+
+    /**
+     *  For browser: credentials_provider is Type AWSCredentials
```
bretambrose, line 130: I don't think we want this kind of a context-sensitive type difference.

lib/common/auth.ts
```diff
@@ -0,0 +1,142 @@
...
+     *  For native:  Type _crt_native_._AwsCredentialsProvider_
+     */
+    credentials_provider?: any;
+
+    /** callback for refresh credential on creating the wws url
+     *  @param credentials_provider. The callback will pass the provider as a parameter
+     */
+    updateCredentialCallback?: Function;
```
bretambrose, line 138: For external credentials sourcing, whether its browser or native, let's just define an interface that returns credentials and invoke it every time we do the signing.  If someone wants to add caching in between, they can.

bretambrose (commented):
The more I look through the more I feel like we should change direction here.

We should unify/expose CredentialsProvider and a browser impl of StaticCredentialsProvider.  When the user passes in fixed credentials the browser code  will create and use the StaticCredentialsProvider, otherwise it'll use the supplied CredentialsProvider, if there is one.  Signing a handshake will always invoke the provider to fetch credentials.  If someone wants caching, they can inject that themselves for now.

lib/browser/aws_iot.ts
```diff
@@ -196,16 +194,31 @@ export class AwsIotMqttConnectionConfigBuilder {
...
-     *
+     * @param customer_provider credential provider used to update credential when session expired (optional)
+     * @param updateCredentialCallback callback used to update credential when session expired (optional)
+     * @param expiration_in_seconds the credential expiration time. The update callback will get called every <expiration_in_seconds> seconds (optional)
+     * 
      * @returns this builder object
      */
-    with_credentials(aws_region: string, aws_access_id: string, aws_secret_key: string, aws_sts_token?: string) {
-        this.params.credentials = {
+     with_credentials(aws_region: string, aws_access_id: string, aws_secret_key: string, aws_sts_token?: string,
```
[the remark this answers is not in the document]
xiazhvera: I will separate api for fixed credential and provider. However, the provider will still need the initial credential as a workaround. mqtt.js does not support promise for transformWsUrl, where we signing the url. This means we need a better way to verify that the credential is ready on the initial connection. Currently I prompt the user to pass the initial credential to establish the initial connection. Then we use provider to update the credentials afterwards.

TwistedTwigleg (approved):

lib/browser/mqtt.ts
```diff
@@ -113,10 +113,7 @@ export interface MqttConnectionConfig {
     password?: string;
 
     /** Options for the underlying websocket connection */
-    websocket?: WebsocketOptions;
-
-    /** AWS credentials, which will be used to sign the websocket request */
-    credentials?: AWSCredentials;
+    websocket: WebsocketOptions;
```
bretambrose, line 116: This seems like a breaking change.

lib/native/auth.ts
```diff
@@ -189,9 +177,6 @@ export interface AwsSigningConfig {
 
     /** Controls what header, if any, should be added to the request, containing the body value */
     signed_body_header?: AwsSignedBodyHeaderType;
-
-    /** Query param signing only: how long the pre-signed URL is valid for */
-    expiration_in_seconds?: number;
```
bretambrose, line 194: Was this intentional?

lib/browser/ws.ts
```diff
@@ -18,19 +19,21 @@ import * as Crypto from "crypto-js";
...
  * @category MQTT
  */
-export interface WebsocketOptions {
+export interface WebsocketOptions extends WebsocketOptionsBase{
     /** Additional headers to add */
     headers?: { [index: string]: string };
     /** Websocket protocol, used during Upgrade */
     protocol?: string;
+
+    credentials_provider?: AWSCredentialsProviderCached;
```
bretambrose, line 28: I feel like this better belongs on the Aws Iot config

lib/browser/ws.ts
```diff
@@ -50,42 +53,48 @@ function make_signing_key(credentials: AWSCredentials, day: string, service_name
...
     const signed_headers = 'host';
+    const service = signing_config.service!;
     const canonical_headers = `host:${url.hostname.toLowerCase()}\n`;
     const payload_hash = Crypto.SHA256(payload, { asBytes: true });
     const canonical_params = url.search.replace(new RegExp('^\\?'), '');
     const canonical_request = `${method}\n${url.pathname}\n${canonical_params}\n${canonical_headers}\n${signed_headers}\n${payload_hash}`;
     const canonical_request_hash = Crypto.SHA256(canonical_request, { asBytes: true });
-    const signature_raw = `AWS4-HMAC-SHA256\n${time}\n${day}/${credentials.aws_region}/${service_name}/aws4_request\n${canonical_request_hash}`;
-    const signing_key = make_signing_key(credentials, day, service_name);
+    const signature_raw = `AWS4-HMAC-SHA256\n${time}\n${day}/${signing_config.credentials.aws_region}/${service}/aws4_request\n${canonical_request_hash}`;
```
bretambrose, line 67: While we may not be able to remove region from credentials due to comaptibility, I think we should be using the signing configs region with a shim that sets it to the region in the credentials, if one exists.

lib/browser/auth.ts
```diff
@@ -11,17 +11,305 @@
...
+        else if (options instanceof AWSCognitoCredentialOptions)
+        {
+            provider = new AWSCognitoCredentialsProvider(options, expire_interval_in_ms);
+        }
+        this.source_provider = provider;
+        this.cached_credentials = this.source_provider?.getCredentials();
+        this.expire_time = this.source_provider?.expire_time;
+    }
+
+    add_provider(provider: CredentialsProvider)
```
bretambrose, line 269: not sure what's going on here

lib/native/aws_iot.ts
```diff
@@ -50,14 +43,14 @@ export interface WebsocketConfig {
...
-    service?: string;
-
     /** (Optional)  TLS configuration to use when establishing the connection */
     tls_ctx_options?: TlsContextOptions;
+
+    /**
+     *  For browser: credentials_provider is Type AWSCredentials
+     *  For native:  Type _crt_native_._AwsCredentialsProvider_
+     */
+    credentials_provider?: AwsCredentialsProvider;
```
bretambrose, line 53: This was formerly  a required parameter

lib/browser/auth.ts
```diff
@@ -11,17 +11,305 @@
  * @preferred
  */
 
+import * as AWS from "aws-sdk";
```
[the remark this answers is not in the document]
xiazhvera: I have the CognitoCredentialsProvider as a build-in class, that's why we need aws-sdk to access the Cognito Credentials Provider features.  We can also move it to the samples, and have the user to implement the provider them selves there. In that case we can remove the dependency on aws-sdk.

lib/browser/auth.ts
xiazhvera: This function allow us to add multiple providers. The cached provider will keep a chain of the credentialsprovider and retrieved credentials from the provider chain, if the cached credentials expired.

lib/browser/ws.ts
xiazhvera: The signing_config is generated based on provider in `create_websocket_url()`. The `sign_url()` will simply use whatever passed in as parameter. We will generate a fresh signing_config every time we call `sign_url`.

lib/native/auth.ts
bretambrose: That's a different expiration time.  This is something that gets written into the sigv4 signing process and has nothing to do with AWS Credentials expiration.

lib/browser/auth.ts
bretambrose: Caches and chain are two separate concerns; we should stick to one.

lib/browser/auth.ts
```diff
@@ -11,17 +11,130 @@
...
+    {
+        return {
+            aws_region: this.options.aws_region,
+            aws_access_id : this.options.aws_access_id,
+            aws_secret_key: this.options.aws_secret_key,
+            aws_sts_token: this.options.aws_sts_token
+        }
+    }
+
+    isExpired = () : boolean =>
```
bretambrose, line 124: If these functions stay, these overrides aren't necessary

lib/browser/aws_iot.ts
```diff
@@ -201,21 +203,31 @@ export class AwsIotMqttConnectionConfigBuilder {
...
+
+    /**
+     * Configures credentials provider (currently support for AWS Cognito Credential Provider) for this connection
+     * @param customer_provider credential provider used to update credential when session expired (optional)
+     * 
+     * @returns this builder object
+     */
+    with_credential_provider( customer_provider : CredentialsProvider) {
+        this.params.credentials_provider = customer_provider;
+        setInterval(()=>{ customer_provider.refreshCredential();}, customer_provider.expire_interval_in_ms);
```
bretambrose, line 230: Mirroring the above comment, I think it's best if we left this up to the provider subclass.  As it currently stands, every connection you use the provider with will spawn an independent refresher task which isn't what we want.

lib/native/auth.ts
```diff
@@ -98,23 +99,23 @@ export enum AwsSignedBodyHeaderType {
...
-     *
-     * @returns a new credentials provider using default credentials resolution rules
-     */
-    static newDefault(bootstrap: ClientBootstrap | undefined = undefined): AwsCredentialsProvider {
-        return super.newDefault(bootstrap != null ? bootstrap.native_handle() : null);
-    }
+/**
+ * Creates a new default credentials provider to be used internally for AWS credentials resolution:
+ *
+ *   The CRT's default provider chain currently sources in this order:
```
bretambrose, line 105: Feels like this is just a white space change that loses indentation

samples/browser/pub_sub/index.ts
```diff
@@ -13,39 +13,163 @@ function log(msg: string) {
...
+ * AWSCognitoCredentialOptions. The credentials options used to create AWSCongnitoCredentialProvider.
+ *
+ * @category Auth
+ */
+class AWSCognitoCredentialOptions implements auth.CredentialsProviderOptions
+{
+    IdentityPoolId : string;
+    Region: string;
+
+    constructor(IdentityPoolId: string, region: string)
```
bretambrose, line 27: I think you can just put `'private` in front of the parameter names and it will automatically turn them into properties on the class and they'll be initialized properly without explicitly assigning.

samples/browser/pub_sub/index.ts
```diff
@@ -13,39 +13,163 @@ function log(msg: string) {
...
+}
+
+/**
+ * AWSCredentialsProviderCached. The AWSCredentialsProviderCached will be our main provider class.
+ * The AWSCredentialsProviderCached will cached the current credential and expired_time, and stores a list of credentials providers. 
+ *          If the credential is not expired, return the cached credentials.
+ *          If the credential is expired, refresh the credential, and check the next credentialsProvider.
+ * @category Auth
+ */
+ export class AWSCredentialsProviderCached extends auth.CredentialsProvider{
```
bretambrose, line 128: With the right refresh approach, I think we can eliminate this class

samples/browser/pub_sub/index.ts
```diff
@@ -13,39 +13,163 @@ function log(msg: string) {
...
 
-async function connect_websocket(credentials: AWS.CognitoIdentityCredentials) {
+async function connect_websocket(provider: auth.CredentialsProvider) {
     return new Promise<mqtt.MqttClientConnection>((resolve, reject) => {
         let config = iot.AwsIotMqttConnectionConfigBuilder.new_builder_for_websocket()
             .with_clean_session(true)
             .with_client_id(`pub_sub_sample(${new Date()})`)
             .with_endpoint(Config.AWS_IOT_ENDPOINT)
-            .with_credentials(Config.AWS_REGION, credentials.accessKeyId, credentials.secretAccessKey, credentials.sessionToken)
+            /** The following line is a sample of static credential. Please note the static credential will fail when web session expires.*/
```
bretambrose, line 161: Can't these comments be removed now?

bretambrose (commented):
I think we're getting close

lib/browser/mqtt.ts
```diff
@@ -115,8 +115,8 @@ export interface MqttConnectionConfig {
     /** Options for the underlying websocket connection */
     websocket?: WebsocketOptions;
 
-    /** AWS credentials, which will be used to sign the websocket request */
-    credentials?: AWSCredentials;
+    /** Options for the underlying credentianls provider */
```
bretambrose, line 118: This is still a breaking change.  We need to support both possibilities.  If they specify credentials, then use them to make a provider internally, otherwise use the provider.

bretambrose (approved):

PR #320, opened 2022-05-23
Update NPM packages

TwistedTwigleg:
*Description of changes:*

Updates a few NPM packages to later versions. Specifically:

* `minimist` updated to version `1.2.6`+
* `shelljs` updated to version `0.8.5`+
* Older versions of `tar` in dependencies (`cmake-js` primarily) are updated to `4.4.19`+
* `prettier` is locked to version `2.3.2`, which is the last version that supports Typescript 3

This commit uses NPM overrides to update these dependencies, as many of them are dependencies of dependencies. The dependencies themselves we cannot easily update without updating all of them in a major overhaul update.

(*I am not sure why it adds so many new lines to `package-lock.json`...*)

________

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*


bretambrose (commented):
I don't think overrides is the way to solve these issues.  It doesn't seem to be available until npm 8 which is very recent and I'm not sure we can require users to have 8 when our node baseline is 10.

TwistedTwigleg:
It looks like there is a couple of other ways we could potentially look at getting similar functionality, but for older versions of NPM:

The first is using `npm shrinkwrap`, which is supported in NPM 6+. It doesn't look as flexible, but it built in to NPM and should work for our purposes. I will need to experiment to know if it will work for sure though, as the documentation on the NPM side is a little parse. [Documentation](https://docs.npmjs.com/cli/v8/commands/npm-shrinkwrap) and [example link](https://stackoverflow.com/a/17423915).

A second way is to use [NPM Force Resolutions](https://www.npmjs.com/package/npm-force-resolutions). This is probably the most similar to the overrides, but it is developed by a third party with its own set of dependencies. It looks like it does a more post-processing approach for changing the `package-lock.json` file. My biggest concern here is that it's maintained outside of NPM, so at best it may be a temporary solution should it stop working and/or it stop being maintained. [Documentation](https://www.npmjs.com/package/npm-force-resolutions) and [example](https://stackoverflow.com/a/62956076)

I'll keep looking into it, but I wanted to share what I have found so far.

bretambrose (approved):

TwistedTwigleg:

PR #321, opened 2022-06-01
Give docs job write permissions.

graebm:
Our Github workflows don't have write permissions by default anymore, so be explicit.


TwistedTwigleg (approved):

TwistedTwigleg:
Looks good, but it looks like a couple of the CI tests ran into an issue with clean sessions.

I recently merged a PR to update dependencies, which seems to be the only change between the last successful run, but I do not think it would cause an issue? It passed CI when I tested and the issue doesn't seem to be related to the dependencies directly, so I think it is unrelated but I'm not 100% sure.

PR #322, opened 2022-06-07
Browser Reconnect Backoff

xiazhvera:
*Description of changes:*
Update the Browser Reconnection logic to add reconnection backoff. We do it by disable the MQTT.js reconnection features, and create new connection.

*Note: The update based on the changes of PR https://github.com/awslabs/aws-crt-nodejs/pull/319 so that the credentials could get updated on reconnection. 


source/mqtt_client_connection.c
```diff
@@ -647,15 +693,15 @@ napi_value aws_napi_mqtt_client_connection_connect(napi_env env, napi_callback_i
     AWS_ZERO_STRUCT(server_name);
     struct connect_args *on_connect_args = NULL;
 
-    napi_value node_args[10];
+    napi_value node_args[12];
```
bretambrose, line 696: This change seems wrong.  The arguments were already passed in new()

lib/browser/mqtt.ts
```diff
@@ -335,12 +368,20 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
         } else {
+            /** Reset reconnect times after reconnect succeed. */
+            this.reset_reconnect_times();
             this.emit('resume', 0, session_present);
         }
     }
 
     private on_offline = () => {
         this.emit('interrupt', -1);
+        const waitTime = this.get_reconnect_time_sec();
```
bretambrose, line 379: I can't seem to find anywhere that increases reconnect_count

lib/browser/mqtt.ts
```diff
@@ -383,6 +430,25 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
+        }
+
+        // Uses "FullJitter" backoff algorithm, described here:
+        // https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+        // We slightly vary the algorithm described on the page,
+        // which takes (base,cap) and may result in 0.
+        // But we take (min,max) as parameters, and don't don't allow results less than min.
+        const cap = this.reconnect_max_sec - this.reconnect_min_sec;
+        const base = Math.max(this.reconnect_min_sec, 1);
+        const sleep = Math.random() * Math.min(cap, base * (2 ** this.reconnect_count));
```
bretambrose, line 448: Is JS math unbounded?  If reconnect_count gets large due to extended connectivity failures, what happens?

samples/browser/pub_sub/index.ts
```diff
@@ -211,10 +212,13 @@ async function main() {
         })
         .then((subscription) => {
             log(`start publish`)
-            connection.publish(subscription.topic, 'NOTICE ME', subscription.qos);
+            var msg_count = 0;
+            connection.publish(subscription.topic, `NOTICE ME {${msg_count}}`, subscription.qos);
```
bretambrose, line 216: technically msg_count is always 0 here

source/mqtt_client_connection.c
```diff
@@ -441,6 +441,52 @@ napi_value aws_napi_mqtt_client_connection_new(napi_env env, napi_callback_info
         }
     }
 
+    /* set reconnect min/max times. beware that user that might have passed only one of these values */
+    napi_value node_reconnect_min_sec = *arg++;
+    napi_value node_reconnect_max_sec = *arg++;
+    if (!aws_napi_is_null_or_undefined(env, node_reconnect_min_sec) ||
+        !aws_napi_is_null_or_undefined(env, node_reconnect_max_sec)) {
+
+        int64_t reconnect_min_sec = 1;
```
bretambrose, line 450: If these are the internal (native) defaults, consider putting named constants in a shared part of the CRT and then using them as the defaults in the configuration objects.  Then these parameters should always be non-null and you can remove the constants here.

samples/browser/pub_sub/index.ts
xiazhvera: The msg_count will be update in the setInterval callback.

lib/browser/mqtt.ts
xiazhvera: If the number is out of bound, JS will return definition "INFINITY" which will be treat as a infinite large number here. The function will work fine. 
Change ** operator to Math.pow though, since IE does not support ** operator.

lib/browser/mqtt.ts
```diff
@@ -100,6 +102,20 @@ export interface MqttConnectionConfig {
      */
     protocol_operation_timeout?: number;
 
+    /**
+     * Minimum seconds to wait between reconnect attempts.
+     * Must be <= {@link reconnect_max_sec}.
+     * Wait starts at min and doubles with each attempt until max is reached.
+     */
+    reconnect_min_sec?: number;
```
bretambrose, line 110: The intent of the default discussion from yesterday was to see if we can drop the nullable property here and initialize the values to the defaults.  Since client creation requires this, then the config init is the only place we'd need to initialize to the defaults.

xiazhvera: Javascript does not allow a default in interface, that's why we set up the default value in MqttConnection.

source/mqtt_client_connection.c
```diff
@@ -441,6 +441,49 @@ napi_value aws_napi_mqtt_client_connection_new(napi_env env, napi_callback_info
         }
     }
 
+    /* set reconnect min/max times. beware that user that might have passed only one of these values */
+    napi_value node_reconnect_min_sec = *arg++;
+    napi_value node_reconnect_max_sec = *arg++;
+    if (!aws_napi_is_null_or_undefined(env, node_reconnect_min_sec) &&
```
bretambrose, line 447: Isn't it impossible for these to be null/undefined now?

bretambrose (approved):
Fix n ship: min and max reconnect in the native impl can't be null/undefined anymore so the unpacking can be simplified.

source/mqtt_client_connection.c
xiazhvera: No as the MQTTConnection should set the value to default when call the function. The binding function will also threw an error message in case there are any changes on node side that removed the max/min value.

PR #332, opened 2022-07-06
Typescript upgrade

bretambrose:
* Update typescript to latest
* Update typedoc to almost-latest MV
* Update typedoc config to account for all the breaking changes
* Swap typedoc plugins to modern ones that still work and do approximately what we had plugins doing before.
* Rework how events are documented
* Move some private browser mqtt connection functionality to the bottom of the class definition
* Move some problematic type definitions that used CrtError (which is defined in browser/native) down to browser/native
* Expose checksums as an export
* Fix some messy type errors due to stricter checking in 4.x
* Update sample package.jsons

### DIFF
--- a/lib/common/auth.ts
+++ b/lib/common/auth.ts
@@ -4,8 +4,6 @@
  */
 
 /**
- * TestCommon
- *
  * @packageDocumentation
  * @module auth
  */
--- a/lib/common/auth.ts
+++ b/lib/common/auth.ts
@@ -4,11 +4,17 @@
  */
 
 /**
+ * TestCommon
+ *
+ * @packageDocumentation
+ * @module auth
+ */
+
+/**
  * Configuration for use in AWS-related signing.
  * AwsSigningConfig is immutable.
  * It is good practice to use a new config for each signature, or the date might get too old.
  *
- * @internal
  */
 export interface AwsSigningConfigBase {
 
@@ -28,7 +34,6 @@ export interface AwsSigningConfigBase {
  * Configuration for websocket signing
  * It is good practice to use a new config for each signature, or the date might get too old.
  *
- * @Internal
  */
 export interface WebsocketOptionsBase {
     /**
--- a/lib/native/aws_iot.ts
+++ b/lib/native/aws_iot.ts
@@ -18,6 +18,7 @@ import { TlsContextOptions } from "./io";
 import * as platform from '../common/platform';
 import { HttpProxyOptions } from "./http";
 import { WebsocketOptionsBase } from "../common/auth";
+import { CrtError } from "./error";
 
 import {
     aws_sign_request,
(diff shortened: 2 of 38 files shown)


lib/native/mqtt.ts
```diff
@@ -256,63 +277,57 @@ export class MqttClientConnection extends NativeResourceMixin(BufferedEventEmitt
...
     /**
      * Emitted when the connection reconnects (after an interrupt). Only triggers on connections after the initial one.
      *
-     * @param event the type of event (resume)
-     * @param listener the event listener to use
-     *
      * @event
      */
-    on(event: 'resume', listener: MqttConnectionResumed): this;
+    static RESUME = 'resume';
```
TwistedTwigleg, line 312: Question: Should we have the `on(event: 'resume', listener: MqttConnectionResumed): this;` code be `on(event: RESUME, listener: MqttConnectionResumed): this;` so it uses this variable?
I'm not sure if Typescript allows it though. It is a super minor thing regardless and so I think it's fine either way.

lib/native/aws_iot.ts
```diff
@@ -178,7 +179,11 @@ export class AwsIotMqttConnectionConfigBuilder {
                     await aws_sign_request(request, signing_config as AwsSigningConfig);
                     done();
                 } catch (error) {
-                    done(error);
+                    if (error instanceof CrtError) {
+                        done(error.error_code);
+                    } else {
+                        done(3); /* TODO: AWS_ERROR_UNKNOWN */
```
TwistedTwigleg, line 185: Question: Should we remove this TODO?

lib/common/auth.ts
```diff
@@ -3,12 +3,18 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+/**
+ * TestCommon
```
TwistedTwigleg, line 7: Question: Do we need to fill this in?

TwistedTwigleg (approved):
Looks good! There was just a few minor things I have questions on, but nothing at all major nor blocking 👍

lib/native/mqtt.ts
bretambrose: I couldn't get that to compile unfortunately.

lib/native/aws_iot.ts
bretambrose: Well, I don't want to make modeling native error codes part of this PR but I also don't want to completely "forget" that this is a temporary work-around.

lib/common/auth.ts
bretambrose: Yeah that was leftover from the experimentation phase.

lib/native/mqtt.ts
TwistedTwigleg: Ah okay, in that case, I would say it's fine as is. We can always be adjusted in the future if we find a way around it.

lib/native/aws_iot.ts
TwistedTwigleg: Ah okay. I think for now, it's fine to leave it to get these fixes out.
Maybe I'll look at adding this as part of my on-call. 👍

PR #335, opened 2022-07-12
Move cmake-js and tar to devDeps

trivikr:
*Issue #, if available:*
Refs: https://github.com/awslabs/aws-crt-nodejs/issues/333

*Description of changes:*
Moves cmake-js and tar to devDeps

*Testing:*

The artifact for `aws-crt` is created by running `npm pack` in the workspace.

Before:

```console
$ test-aws-crt> rm -rf * && npm init -y

$ test-aws-crt> npm install ../aws-crt-nodejs/aws-crt-1.0.0-dev.tgz

$ test-aws-crt> du -sh node_modules 
28M     node_modules

$ test-aws-crt> find node_modules -type d -maxdepth 1 | wc -l
118
```

After:

```console
$ test-aws-crt> rm -rf * && npm init -y

$ test-aws-crt> npm install ../aws-crt-nodejs/aws-crt-1.0.0-dev.tgz

$ test-aws-crt> du -sh node_modules 
17M     node_modules

$ test-aws-crt> find node_modules -type d -maxdepth 1 | wc -l
56
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #338, opened 2022-07-13
Move tar outside of dependencies

TwistedTwigleg:
*Description of changes:*

~~Mirror of https://github.com/awslabs/aws-crt-nodejs/pull/335 but merged with latest changes in `main` and the package lock regenerated with npm.~~

Moves `cmake-js` and `tar` outside of dependencies and modifies the build script to download `tar` and `cmake-js`, use it, and store it in a sub-folder for reuse that is completely separate from the main node modules. This prevents `cmake-js` and `tar` from sticking around when it is only needed for building the source and not in the functionality of the CRT itself (and in the case of `tar`, it is needed only for a single download once to get the source).

With this change, `cmake-js` and `tar` is not sticking around in the node dependencies forever when it is only needed for build-time operations.

______________


TwistedTwigleg:
Wait on merging even if CI is successful for now - need to double check things.

scripts/build.js
```diff
@@ -89,13 +93,88 @@ function checkChecksum(url, local_file) {
...
+        console.log("ERROR - npm could not remove " + package_name + "!");
+        throw err;
+    }
+}
+
+
 async function fetchNativeCode(url, version, path) {
+    // Get tar if it doesn't exist
+    var remove_tar_at_end = npmDownloadAndInstallRuntimePackage("tar", tar_version);
+    var tar = require('tar');
```
xiazhvera, line 165: Would we need a try-catch here to make sure the download succeed? While, I'm not sure how would js handle the errors, ignore it if it is not necessary here.

xiazhvera (approved):
Seems good to me! Probably need improve the error check if necessary.

scripts/build.js
TwistedTwigleg: I'm not sure right off. I think it would just stop the script from running, which would in turn prevent the build. However, I should test to find out 👍

TwistedTwigleg: Just checked and it crashes, which is probably fine, but it generates a lot of error messages and stuff that is a little cluttered and hides the real cause of the download failing. Catching the error and exiting with an error code of 1 looks much better and is easier to read.

I adjusted the PR accordingly.

scripts/build.js
```diff
@@ -89,13 +93,95 @@ function checkChecksum(url, local_file) {
...
+        }
+    } catch (error) {}
+
+    // If it is not found in either, then download it into our node_modules
+    try {
+        console.log("Could not find " + package_name);
+        console.log("Downloading " + package_name + " from npm for build...");
+        // Try to intall the given package and ONLY the given package. Will throw an exception if there is an error.
+        if (package_version != null) {
+            child_process.execSync("npm install --no-package-lock --save-dev --ignore-scripts " + package_name + "@" + package_version);
```
graebm, line 134: This seems really sketchy

I realize the whole premise of downloading code and compiling it as part of the install process is already a tough pill to swallow, but is it going to far to install additional NPM dependencies that aren't declared in `package.json`?

Let's talk with someone that has deeper familiarity with what is considered acceptable in Javascript-Land

But if this is acceptable, we could do it with `cmake-js` as well 🤷‍♀️

TwistedTwigleg: I will leave it for now until and see if I can get more feedback to make sure the change is okay with those experienced in Javascript 👍

scripts/build.js
```diff
@@ -89,13 +93,95 @@ function checkChecksum(url, local_file) {
...
+ * @param {*} package_name The name of the package you want to download (example: 'cmake-js')
+ * @param {*} package_version The version of the package to download - leave blank for latest. (example: '6.3.2')
+ * @returns True if the package was downloaded dynamically, otherwise false.
+ */
+function npmDownloadAndInstallRuntimePackage(package_name, package_version=null) {
+    console.log("Looking for " + package_name + " as a dependency...");
+
+    // Do we have it in node modules? If so, then use that!
+    try {
+        if (fs.existsSync("./node_modules/" + package_name)) {
```
kuhe, line 113: you can use `require.resolve` if you want to check for a module and aren't specifically looking in the immediate node_modules directory.

scripts/build.js
kuhe: For good or bad, this doesn't faze me as a JS developer. 

I suggest doing this in a subfolder workspace such as `your_package_location/scripts/`.
You can leave a dummy `package.json` with `private:true` which will allow temporary modules to be installed there if they are not part of the user's workspace. 

That way, the workflow could look like this:

- user runs `npm install aws-crt`
- the `aws-crt` package is downloaded to their node_modules (root)
    - `aws-crt` postinstall tries to use the root `cmake-js` et. al. runtime packages
    - if not present or of the wrong version, download them into a subfolder of `aws-crt` so as not to affect the user's root workspace node_modules / package.json / package-lock.json etc.
    - if not able to download, let the user know they will need to install certain dev dependencies and re-run the `aws-crt` install script.

This would avoid bringing non-runtime packages into both the user's development workspace (root node_modules) as well as the user's potential runtime distribution (package.json->dependencies).

scripts/build_dependencies/build_step_cmake.js
```diff
@@ -0,0 +1,116 @@
...
+module.exports = {
+    cmake: null,
+    cmake_version: "6.3.2",
+
+    /**
+     * Will compile the source code for the CRT using cmake-js. Will automatically either use
+     * or download the runtime-package for cmake-js as needed.
+     */
+    performStep: async function () {
+        if (utils.npmCheckIfPackageExists("cmake-js") == true) {
```
kuhe, line 20: minor item: you don't need `== true`

scripts/build_dependencies/build_utils.js
```diff
@@ -0,0 +1,85 @@
...
+        } catch (error) {}
+
+        // Do we have it in our global list?
+        try {
+            var list_output = child_process.execSync("npm list -g --depth 0 " + package_name, {encoding: "utf8"});
+            if (list_output.indexOf(package_name + "@" + package_version) !== -1) {
+                console.log("Found " + package_name + " in node list!");
+                return true;
+            }
+        } catch (error) {}
```
kuhe, line 27: I'm not sure if there's any effective difference to what you have here, but you can use `require.resolve('package_name')` to check for existence and `require('package_name/package.json').version` to check the version.

Since you will be calling `require` for the package(s), `require.resolve` uses the same lookup.

scripts/build_dependencies/build_step_tar.js
```diff
@@ -0,0 +1,63 @@
...
+
+        process.chdir(__dirname);
+        if (this.tar == null) {
+            try {
+                utils.npmDownloadAndInstallRuntimePackage("tar", tar_version);
+                this.tar = require('tar');
+            } catch (error) {
+                console.log("ERROR: Could not download tar! Cannot build CRT");
+                console.log("Please install tar verion " + this.tar_version + " and then run the aws-crt install script again");
+                process.exit(1);
```
kuhe, line 40: if this exits, would it leave the user in a different directory?

TwistedTwigleg: Ah, good catch! Yeah, I need to revert the directory if this happens. 👍

scripts/build_dependencies/build_utils.js
```diff
@@ -0,0 +1,85 @@
...
+        } catch (err) {
+            console.log("ERROR - npm could not download " + package_name + "! " + package_name + " is required to build the CRT");
+            throw err;
+        }
+    },
+
+    /**
+     * Tells NPM to uninstall a package. This should only be used to clean up a dynamic package downloaded with the
+     * npmDownloadAndInstallRuntimePackage function, as otherwise it could remove a non-dynamic package.
+     * @param {*} package_name The name of the package you want to delete (example 'cmake-js')
```
kuhe, line 73: minor issue: these `{*}` are supposed to be type annotations like `@param {string} package_name`

kuhe (approved):

scripts/build_dependencies/build_utils.js
TwistedTwigleg: I looked at using `require.resolve` and while it works for the check, then trying to use `require` after installing the package leads to a `elifecycle errno 1` crash. It seems that something about using `require.resolve` makes a following `require` call in the same Javascript instance not work. It works fine if you then rerun the script again though, but it doesn't allow for checking, seeing it is not there, downloading the package, and then using it all in one go.

I tried unloading after the `require.resolve` call, but this didn't work either. For now, I'm going to use the current solution since it seems to work but I'll add a TODO note to look at changing this to `require.resolve` in the future.

scripts/build_dependencies/build_step_tar.js
```diff
@@ -0,0 +1,63 @@
...
+
+    tar : null,
+
+    /**
+     * Will download the file at the given url with the given version to the given path using tar.
+     * Will automatically either use or download the runtime-package for tar as needed.
+     */
+    performStep: async function (url, version, path) {
+        if (utils.npmCheckIfPackageExists("tar")) {
+            await this.buildSource(url, version, path);
```
TingDaoK, line 19: I didn't see this function here. Is it missed or did I misunderstand it?

TwistedTwigleg: Yeah, this was a copy-paste error 😅 
I'll see if I can test it locally to make sure it is working as expected 👍

TwistedTwigleg: Tested and confirm it works now 👍

scripts/build_dependencies/build_step_tar.js
```diff
@@ -0,0 +1,63 @@
...
+     * Will download the file at the given url with the given version to the given path using tar.
+     * Will ALWAYS download tar to the node_modules in scripts/build_dependencies/node_modules.
+     */
+    getPackageAndFetchNativeCode : async function (url, version, path) {
+        const workDir = path.join(__dirname, "../../")
+
+        process.chdir(__dirname);
+        if (this.tar == null) {
+            try {
+                utils.npmDownloadAndInstallRuntimePackage("tar", tar_version);
```
TingDaoK, line 35: there is no `tar_version` as well. So, again, make sure it works at least.

scripts/build_dependencies/build_step_axios.js
```diff
@@ -0,0 +1,106 @@
...
+ */
+const fs = require("fs");
+const crypto = require('crypto');
+const utils = require('./build_utils');
+
+module.exports = {
+
+    axios: null,
+    clean_up_axios: false,
+    axios_version: "0.24.0",
```
TingDaoK, line 13: I am not a fan of this hard-coded version... 
- We cannot tell if any CVE or something happens to that package at that version unless manually checking the code, which could be a concern
- If the lib was existed, and it's not this version, we will not know. 

But, I cannot think of a better way😞. 
Does any other nodejs lib using the similar approach to fetch dependency dynamically? How do they deal with the version?

scripts/build_dependencies/build_step_tar.js
TwistedTwigleg: Good catch, I missed that.

I did test it and it works though - it will just pull the latest version of the package if you do not manually specify a version. It just doesn't specify to NPM in the `npm install` call under the hood which version you want if you pass `null` or `undefined`.

scripts/build_dependencies/build_step_axios.js
TwistedTwigleg: I'm not sure, I would need to look around and see.

We could put it in our dev dependencies and then pull it out of the `package.json` that way. Then it would get automatically code scanned by Dependabot.

Though the dependencies here are completely isolated from the rest of the node modules by being in the `node_modules` folder of the `scripts/build_dependencies` folder. So even if there was a vulnerability, it would only be for building the source and not the output or anything a customer/consumer of the CRT would see.
Not saying it is something we shouldn't fix, but the security impact is minimal because it just gets downloaded and used only for a build.

scripts/build_dependencies/build_step_tar.js
TwistedTwigleg: Yeah, unfortunately. Typescript will, but not Javascript. It's part of what makes it *really* hard to debug at times, from past experience.

scripts/build_dependencies/build_step_axios.js
TingDaoK: I am not totally against it, but if there is a better way, we should take it instead.

TingDaoK (approved):

scripts/build_dependencies/build_step_tar.js
TwistedTwigleg: Okay - I did a bunch more tests and this was a good catch. I'm super glad you asked me to check because I probably would have missed it if I was not taking another careful look!

It was sort of working, but it was due to a bunch of other errors that basically made it not fully go down the code path! There was also some other cases there the version wasn't being used and the tar extract was pointing to the wrong place as well... So it worked, because I was running it in the CRT and already had the source code downloaded and in the right location, but if it was outside of that environment, it wouldn't have worked.

So... All of that is fixed now! I fully tested it and it all works and I made sure there was not a single error in the NPM build log. I missed it was doing some wonky things. Sigh - Javascript is *fun*.

scripts/build_dependencies/build_step_axios.js
TwistedTwigleg: I did some looking, and it looks like the general consensus is "don't do it" based on other Javascript packages and such. I got around this by putting it in the `devDependencies` and then loading the `package.json` and getting the version number from there. Now we have the advantage of getting Dependabot code scanning, can easily change the version in a single place, and it is more in line with what I have seen in other Javascript packages/modules.

TwistedTwigleg:
Thanks for the reviews everyone! Merging into `main`...

PR #351, opened 2022-08-30
fix our release pipeline

TingDaoK:
*Issue #, if available:*

- Our pack stage of the release pipeline doesn't have Cmake, as it should not need to rebuild the native addon
- However the prepare script will always rebuild the native addon, which I doubt it's needed. At least for our release pipeline.
- Did a quick search about how people did for other nodejs lib with native addon. 
   - Some use prepare to build, but they don't have install eg: [react-pdf](https://github.com/diegomura/react-pdf/blob/master/package.json). So that if people run [npm-pack](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-pack) or publish, they will have the native addon available without install.
   - Some use install to build, and the prepare just do the tsc thing. eg: [Couchbase](https://github.com/couchbase/couchnode/blob/master/package.json). User will need to run npm install first in this case.

*Description of changes:*

- I think we should skip the build of native addon for the prepare script, maybe it's nice to have native addon available without `npm install`? I don't have strong feeling about it.
- Well, the other end, the reason for pack and publish is for us to publish the package, and we don't really want to just publish one prebuild binary. So, I don't think it really matters.
- Fix the lie about `--rebuild`

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:browser": "npx jest --runInBand --verbose --config test/browser/jest.config.js --forceExit",
     "test:browser:ci": "npm run install:puppeteer && npm run test:browser",
     "install:puppeteer": "npm install --save-dev jest-puppeteer puppeteer @types/puppeteer",
-    "prepare": "node ./scripts/tsc.js && node ./scripts/install.js",
+    "prepare": "node ./scripts/tsc.js",
     "install": "node ./scripts/install.js"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:browser": "npx jest --runInBand --verbose --config test/browser/jest.config.js --forceExit",
     "test:browser:ci": "npm run install:puppeteer && npm run test:browser",
     "install:puppeteer": "npm install --save-dev jest-puppeteer puppeteer @types/puppeteer",
-    "prepare": "node ./scripts/tsc.js",
+    "prepare": "node ./scripts/tsc.js && node ./scripts/install.js",
     "install": "node ./scripts/install.js"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
     "test:browser": "npx jest --runInBand --verbose --config test/browser/jest.config.js --forceExit",
     "test:browser:ci": "npm run install:puppeteer && npm run test:browser",
     "install:puppeteer": "npm install --save-dev jest-puppeteer puppeteer @types/puppeteer",
-    "prepare": "node ./scripts/tsc.js && node ./scripts/build.js",
-    "install": "node ./scripts/install.js --production"
+    "prepare": "node ./scripts/tsc.js",
+    "install": "node ./scripts/install.js"
   },
   "devDependencies": {
     "@types/crypto-js": "^3.1.43",
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -7,10 +7,13 @@ const process = require("process");
 const path = require("path");
 const fs = require("fs");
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -22,8 +22,8 @@
     "test:browser": "npx jest --runInBand --verbose --config test/browser/jest.config.js --forceExit",
     "test:browser:ci": "npm run install:puppeteer && npm run test:browser",
     "install:puppeteer": "npm install --save-dev jest-puppeteer puppeteer @types/puppeteer",
-    "prepare": "node ./scripts/tsc.js && node ./scripts/build.js",
-    "install": "node ./scripts/install.js --production"
+    "prepare": "node ./scripts/tsc.js",
+    "install": "node ./scripts/install.js"
```
graebm, line 26: I can't figure out why that `--production` flag was ever there ???

I see it was added in [PR #329](https://github.com/awslabs/aws-crt-nodejs/pull/329/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R23)

but why?

scripts/install.js
```diff
@@ -7,10 +7,13 @@ const process = require("process");
...
 const fs = require("fs");
 
-const binaryDir = path.join('dist', 'bin', `${os.platform()}-${os.arch()}`, 'aws-crt-nodejs.node');
-if (fs.existsSync(binaryDir)) {
-    // Don't continue if the binding already exists (unless --rebuild is specified)
-    process.exit(0);
+if (!process.argv.includes('--rebuild')) {
+    const binaryDir = path.join('dist', 'bin', `${os.platform()}-${os.arch()}`, 'aws-crt-nodejs.node');
+    if (fs.existsSync(binaryDir)) {
+        // Don't continue if the binding already exists (unless --rebuild is specified)
```
graebm, line 13: yay, now this pre-existing comment is correct. good work

graebm (approved):

graebm (changes requested):
wait ... I'm reading docs and worried that the `prepare` script is bad

package.json
TingDaoK: No idea, but it is not doing anything and makes me confused. So, removed it

graebm (approved):

package.json
TingDaoK: The `install.js` and `build.js` is all our code, and none of those uses `--production` anywhere.

package.json
```diff
@@ -22,8 +22,8 @@
     "test:browser": "npx jest --runInBand --verbose --config test/browser/jest.config.js --forceExit",
     "test:browser:ci": "npm run install:puppeteer && npm run test:browser",
     "install:puppeteer": "npm install --save-dev jest-puppeteer puppeteer @types/puppeteer",
-    "prepare": "node ./scripts/tsc.js && node ./scripts/build.js",
```
DmitriyMusatkin, line 25: prepare runs both during packaging and when user installs on the system. we probably dont need it for packing, but wouldnt we want to have it for when user install and binaries are not available for their system?

TingDaoK: User will install the package from [npm install](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-install), which will run the install script and build the native addon if not existed. 

for npm publish and npm pack, I don't believe our user should ever use those.

But, after double check the documentation, maybe we want to support https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-cache-add?

package.json
DmitriyMusatkin: its a process wide argv, so something outiside of our code might have been picking it, but i doubt it. looks unused

package.json
graebm: in the failing release pipeline, why isn't install.js just skipping the cmake build? Doesn't it just skip building if the binary is already there?

graebm (approved):

PR #356, opened 2022-09-15
Cognito provider

bretambrose:
* Adds a binding to the native Cognito credentials provider implementation in the node build


lib/native/auth.spec.ts
```diff
@@ -5,8 +5,9 @@
 
 import { auth as native, http as native_http } from '../index';
 import { io as native_io } from '../index';
+import { ProxyConfig, ProxyTestType } from "@test/proxy";
 
-import { InputStream } from './io';
+import { InputStream} from './io';
```
TingDaoK, line 10: super trivial: add the space back?

lib/native/auth.spec.ts
```diff
@@ -187,4 +188,104 @@ test('Default credentials provider create no bootstrap', async () => {
...
+    };
+
+    const credentials_provider = native.AwsCredentialsProvider.newCognito(config);
+
+    expect(credentials_provider);
+
+    const signing_result = await do_body_request_signing(credentials_provider);
+
+    expect(signing_result.method).toBe(SIGV4TEST_METHOD);
+    expect(signing_result.path).toBe(SIGV4TEST_PATH);
```
TingDaoK, line 290: I believe this test never runs on CI. The only place we set up the proxy config are the codebuild job, which simply runs `npm test` [here](https://github.com/awslabs/aws-crt-nodejs/blob/main/codebuild/linux-integration-tests.sh). So, it will not go through the builder to setup the cognito config.

Maybe update the codebuild job to use builder instead?

source/auth.c
```diff
@@ -187,6 +203,250 @@ static napi_value s_creds_provider_new_static(napi_env env, const struct aws_nap
...
+
+            if (AWS_NGNPR_VALID_VALUE != aws_napi_get_named_property_as_bytebuf(
+                                             env,
+                                             napi_token_pair,
+                                             AWS_NAPI_KEY_IDENTITY_PROVIDER_TOKEN,
+                                             napi_string,
+                                             &identity_provider_token_buf)) {
+                AWS_LOGF_ERROR(
+                    AWS_LS_NODEJS_CRT_GENERAL,
+                    "s_aws_cognito_credentials_provider_config_init - required property 'identityProviderToken' "
```
TingDaoK, line 322: hmmm, I'd like to learn the reason about log out the function name here. My personal feeling is that the function name would only be helpful to maintainer (? We can always do search on the log info to locate the code). It's not helpful for customer?

source/auth.c
```diff
@@ -49,6 +58,13 @@ napi_status aws_napi_auth_bind(napi_env env, napi_value exports) {
             .arg_types = {napi_string, napi_string, napi_string},
             .attributes = napi_static,
         },
+        {
+            .name = "newCognito",
+            .method = s_creds_provider_new_cognito,
+            .num_arguments = 4,
+            .arg_types = {napi_undefined, napi_undefined, napi_undefined, napi_undefined},
```
TingDaoK, line 65: Hmmm, we used all the args as `napi_undefined`, but we read some of those as `napi_external` [here](https://github.com/awslabs/aws-crt-nodejs/pull/356/files#diff-ff0326d36d3ec5896c557f1325368efd5a0c9281c31c1969816597d96ea8e806R425). I did a quick search but not seeing anything useful about the difference of the types here... Just pointing out as I feel confusing.

lib/native/auth.spec.ts
bretambrose: I've updated the environment setup script in s3 to also initialize the cognito environment variable.

source/auth.c
bretambrose: I may be misinterpreting, but I don't agree with the notion that "only information useful to a customer should go in logging"

source/auth.c
bretambrose: undefined means don't enforce any type checking if I'm understanding Colden's original implementation; just following the pattern of earlier bindings here.

source/auth.c
```diff
@@ -187,6 +203,250 @@ static napi_value s_creds_provider_new_static(napi_env env, const struct aws_nap
...
+    if (AWS_NGNPR_INVALID_VALUE ==
+        aws_napi_get_named_property_as_bytebuf(
+            env, node_config, AWS_NAPI_KEY_CUSTOM_ROLE_ARN, napi_string, &config->custom_role_arn)) {
+        AWS_LOGF_ERROR(
+            AWS_LS_NODEJS_CRT_GENERAL,
+            "s_aws_cognito_credentials_provider_config_init - required property 'identity' could not be extracted from "
+            "config");
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        goto done;
+    }
```
TingDaoK, line 364: The log info is `identity`, which doesn't match `AWS_NAPI_KEY_CUSTOM_ROLE_ARN`.

source/auth.c
TingDaoK: Yeah, but I also don't get how the name of the static function helps maintainer as well. 

Also, the details about which field or part from the Nodejs `CognitoCredentialsProviderConfig` are logged in C. The nodejs exception only has `Failed to initialize cognito provider configuration from node config`.

From the experience with Java SDK, I think we should try to have those debug info throw out to the language level. Like it's not a good experience to let nodejs user enable the C log and check the message out. 

So, I think maybe we can move those check of the configurations to nodejs level instead? 

On the other hand, I think yeah, maybe the `Failed to initialize cognito provider configuration from node config` is good enough for people to know to check their configuration and the document to get the idea. 🤷‍♂️

TingDaoK (approved):
Still have thoughts about the logs, but other than that everything looks good.
Discuss(?) & ship

lib/native/auth.spec.ts
TingDaoK: Looks like it breaks all the ci that has not updated the builder version...
https://github.com/awslabs/aws-c-http/actions/runs/3101997039/jobs/5023883935#step:2:4009

source/auth.c
bretambrose: Moving all the validation into the CRTs implies we're duplicating that validation across all CRTs rather than having it in one spot, which kind of goes against one of our guiding principles.

lib/native/auth.spec.ts
bretambrose: That seems like a problem with the stuff invoking the downstream builds.  Given builder versioning, you shouldn't be doing a downstream build of a target using *your* builder version, you should be using the downstream target's builder version.

source/auth.c
bretambrose: Logging's primary purpose is to help us debug.  I find the function name helpful because it's an instant zero-in search term.  Otherwise you have to muddle around trying to figure out what is a sufficient search string.  Sometimes that's easy, sometimes it's a real pain, it all depends on the wording in the line (which is often fragmented by substitution parameters which disrupt the search).

PR #361, opened 2022-10-24
Added unit test to setup Mqtt Connect through EC-based certificate and key

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/common/mqtt.spec.ts
+++ b/lib/common/mqtt.spec.ts
@@ -52,6 +52,44 @@ test('MQTT Connect/Disconnect', async () => {
     await expect(promise).resolves.toBeTruthy();
 });
 
+test('MQTT ECC key Connect/Disconnect', async () => {
+    const promise = new Promise(async (resolve, reject) => {
+        let aws_opts: Config;
+        try {
+            aws_opts = await fetch_credentials();
+        } catch (err) {
+            reject(err);
+            return;
+        }
+
+        const config = AwsIotMqttConnectionConfigBuilder.new_mtls_builder(aws_opts.ecc_certificate, aws_opts.ecc_private_key)
+            .with_clean_session(true)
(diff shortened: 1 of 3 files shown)


lib/common/mqtt.spec.ts
```diff
@@ -52,6 +52,44 @@ test('MQTT Connect/Disconnect', async () => {
...
+    const promise = new Promise(async (resolve, reject) => {
+        let aws_opts: Config;
+        try {
+            aws_opts = await fetch_credentials();
+        } catch (err) {
+            reject(err);
+            return;
+        }
+
+        const config = AwsIotMqttConnectionConfigBuilder.new_mtls_builder(aws_opts.ecc_certificate, aws_opts.ecc_private_key)
```
graebm, line 65: in aws-crt-nodejs, there are actually 2 implementations of everything: 1 for native and 1 for browser

tests in `lib/common/*` are for tests that will run in both native and browser

I was surprised to see that the new ECC tests worked in browser! (I'm not being sarcastic, I really am a newb on security stuff and thought this was a nice surprise)

looking into it ... sadly ... I discovered that [new_mtls_builder()](https://github.com/awslabs/aws-crt-nodejs/blob/380028fbf21e27fa735f02b460b882a41a52a424/lib/browser/aws_iot.ts#L45-L52) in the browser results in you being lied to:
```
    /**
     * For API compatibility with the native version. Does not set up mTLS.
     *
     * @returns a new websocket connection builder object with default TLS configuration
     */
    static new_mtls_builder(...args: any[]) {
        return AwsIotMqttConnectionConfigBuilder.new_builder_for_websocket();
    }
```

So this is not really doing mutual TLS, it's doing WebSockets with a SigV4 HTTP handshake.

Anyway ... solution is to move this test to `lib/native/mqtt.spec.ts`

graebm (approved):

PR #368, opened 2022-11-04
Handle expired STS token

richarddd:
*Description of changes:*
async call to connect doesn't complete if credentials have expired. If token is expired the `MqttClient` will emit `end` event which currently not handled by the connection callbacks.

This PR also listens for end events and thus the promise completes with an error if the token is expired.


richarddd:
Fixed now!

PR #408, opened 2023-02-06
MQTT311 extra connection callbacks

TwistedTwigleg:
*Description of changes:*

Adds additional connection callbacks to MQTT311. Specifically, adds `onConnectionSuccess`, `onConnectionFailure`, and `onConnectionClosed`. These callbacks mirror those in V1 and were added to both NodeJS (desktop) and the browser..

_____________


lib/browser/mqtt.ts
```diff
@@ -56,6 +56,34 @@ export type MqttConnectionError = (error: CrtError) => void;
...
 
+/**
+ * Listener signature for event emitted from an {@link MqttClientConnection} when the connection has been
+ * connected successfully.
+ *
+ * @param session_present A boolean indicating if the connection resumed a session
+ *
+ * @category MQTT
+ */
+export type MqttConnectionSuccess = (session_present: boolean) => void;
```
TwistedTwigleg, line 67: Note: MQTT.js doesn't provide a return code from the server upon a successfully connect, so we do not expose it here. NodeJS/Desktop does expose it, and so we return it there, but the browser doesn't really have an equivalent.

bretambrose (commented):
Would really like to see tests verifying the expected callback behavior.

lib/browser/mqtt.ts
```diff
@@ -548,26 +620,46 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
         /* Only try and reconnect if our desired state is connected, ie no one has called disconnect() */
         if (this.desiredState == MqttBrowserClientState.Connected) {
             const waitTime = this.get_reconnect_time_sec();
             this.reconnectTask = setTimeout(() => {
-                    /** Emit reconnect after backoff time */
-                    this.reconnect_count++;
-                    this.connection.reconnect();
-                },
+                /** Emit reconnect after backoff time */
+                this.reconnect_count++;
```
bretambrose, line 635: Preferred the previous whitespace layout tbh

lib/browser/mqtt.ts
```diff
@@ -548,26 +620,46 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
+            let closedCallbackData = {} as OnConnectionClosedResult;
+            this.emit("closed", closedCallbackData);
+        }
     }
 
     private on_error = (error: Error) => {
         this.emit('error', new CrtError(error))
+
+        // If we were trying to connect but got an error, then it's a connection failure!
+        if (this.desiredState == MqttBrowserClientState.Connected) {
```
bretambrose, line 659: I'm not sure about this.  Connection failure is emitted in reaction to on_close in the mqtt5 client.

lib/browser/mqtt.ts
TwistedTwigleg: I just ran it through the default VSCode formatter. I can look at adjusting it though back to where it was 👍

lib/browser/mqtt.ts
TwistedTwigleg: It's the only way you can know whether a connection failed due to a ConnACK and we listen for it to determine if the connection failed with the connection promise. MQTT.js describes it as "Emitted when the client cannot connect (i.e. connack rc != 0) or when a parsing error occurs.".

That said though, I did move the code to where we wait for the connection to possibly fail and return the failure to the connection promise, which feels more correct and a better spot to emit the event.

lib/browser/mqtt.ts
```diff
@@ -390,13 +456,51 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
+                reject(crtError);
             };
+            this.connection.once('error', on_connect_error);
+
+            /**
+             * Annoyingly, MQTT.JS does NOT return socket errors or connection failures through websockets on the browser! This
+             * is because the 'error' result from the websocket 'stream' is not bubbled up and so we have no idea we failed to connect.
+             * This means we have to catch this error manually and send it up ourselves to catch the error.
+             */
+            const on_stream_connect_error = (error: Error) => {
```
TwistedTwigleg, line 471: Annoying, this is the only way to get whether the connection actually failed with MQTT.js websockets. It is not perfect, but it's better than nothing at all.

References:
* https://github.com/mqttjs/MQTT.js/issues/876
* https://stackoverflow.com/questions/49413553/mqtt-js-how-to-handle-connection-errors
* https://github.com/mqttjs/MQTT.js/issues/1282

Note that this will only catch the first failed connection, subsequent failed re-connection attempts will not be caught.

bretambrose: I don't agree.  I just checked the 5 client which only uses error to store the particular error.  It does not emit anything; it only emits on closed events.  The closed events are 1-1 with actual connection failures.

TwistedTwigleg: I did some digging and you are right! The `closed` event is the one that needs to be followed, NOT the `error` event (despite MQTT.js saying it's called with ConnAck packets). I did some looking into how MQTT5 does it and realized I was looking in the wrong place (`error`) and was using the wrong states when I was trying to do it in the right place (`close`) when experimenting.

In other words: I was expecting `error` to be invoked when there was a problem and not `close`, and then when I tried to mimic MQTT5 locally, I had the state checking wrong...

Anyway - it is fixed now! Now it checks to see if a connection was called and will emit the connection failure accordingly based on `close` and whether the client intended to connect. I did a test with MQTT5 and compared it to this, and now this PR emits like MQTT5.

TwistedTwigleg:
This PR now also removes the `unacked` operation statistics checks because they are **really** flaky in CI due to being heavily dependent on the socket speed of the container. We can add them back later if we find a good way to test them.

xiazhvera (approved):

TwistedTwigleg:
Thanks! Merging into main...

PR #414, opened 2023-02-24
More usefull errors when missing data

TwistedTwigleg:
*Issue #, if available:*

Fixes https://github.com/aws/aws-iot-device-sdk-js-v2/issues/93

*Description of changes:*

Adjusts the Typescript code to check if arguments are `null` or `undefined` IF their properties or methods are being accessed and the argument is non-optional. While this should never occur, and we have checks in the C code to make sure the arguments are non-null, we have several cases where we take an argument, call something like `example.native_handle()` but do not check if `native_handle` is defined or not before doing so. For Typescript, this is not an issue because Typescript will not allow you to pass `null` or `undefined`, but the NodeJS CRT (and extension IoT SDK) are used by those with Javascript, which does not have this typing guarantee.

This can lead to customers getting issues like `TypeError: Cannot read property 'native_handle' of undefined` when they miss a property. This PR changes it so the error returned will be `Error: MqttClientConnection constructor: socket_options in configuration not defined`.

This PR also fixes a number of small typos in the documentation I found while going through all the Typescript code, as well as marking `socket_options` as NOT optional for MQTT connections, as the code does not allow making it `null`/`undefined` (even prior to this PR) and even if we pass `null` to the C code, in aws-c-mqtt we deference the pointer and this would lead to a crash.

__________

### DIFF
--- a/lib/browser/aws_iot.ts
+++ b/lib/browser/aws_iot.ts
@@ -201,19 +201,19 @@ export class AwsIotMqttConnectionConfigBuilder {
     }
 
     /**
-     * Configures Static AWS credentials for this connection. 
+     * Configures Static AWS credentials for this connection.
      * Please note that the static credential will fail when the web session expired.
      * @param aws_region The service region to connect to
      * @param aws_access_id IAM Access ID
      * @param aws_secret_key IAM Secret Key
      * @param aws_sts_token session credentials token (optional)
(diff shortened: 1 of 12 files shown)


xiazhvera (approved):

TwistedTwigleg:

PR #415, opened 2023-02-25
Event stream message

bretambrose:
* Initial message bindings.  Header types will be fully exercised/tested once streams are complete.

### DIFF
--- a/lib/native/eventstream.ts
+++ b/lib/native/eventstream.ts
@@ -73,10 +73,10 @@ export type Payload = string | Record<string, unknown> | ArrayBuffer | ArrayBuff
  */
 const MAX_INT8 : number = 127;
 const MIN_INT8 : number = -128;
-const MAX_INT16 : number = 65535;
-const MIN_INT16 : number = -65536;
-const MAX_INT32 : number = ((1 << 31) - 1);
-const MIN_INT32 : number = -(1 << 31);
+const MAX_INT16 : number = 32767;
+const MIN_INT16 : number = -32768;
+const MAX_INT32 : number = 2147483647;
+const MIN_INT32 : number = -2147483648;
 const MAX_INT64 : bigint = BigInt("9223372036854775807");
 const MIN_INT64 : bigint = BigInt("-9223372036854775808");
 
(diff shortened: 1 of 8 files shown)


lib/native/eventstream.ts
```diff
@@ -73,13 +73,15 @@ export type Payload = string | Record<string, unknown> | ArrayBuffer | ArrayBuff
  */
 const MAX_INT8 : number = 127;
 const MIN_INT8 : number = -128;
-const MAX_INT16 : number = 65535;
-const MIN_INT16 : number = -65536;
-const MAX_INT32 : number = ((1 << 31) - 1);
-const MIN_INT32 : number = -(1 << 31);
+const MAX_INT16 : number = 32767;
```
sbSteveK, line 76: Dang, nice catch.

lib/native/eventstream.ts
```diff
@@ -91,13 +93,21 @@ export class Header {
...
+
     /**
      * Create a new boolean-valued message header
      *
      * @param name name of the header
      * @param value value of the header
      */
     static newBoolean(name: string, value: boolean): Header {
+        if (!Header.isHeaderNameValid(name)) {
+            throw new CrtError(`Event stream header name (${name}) is not valid`);
```
sbSteveK, line 108: Trivial: The CrtError is identical across all these checks, could this be thrown in the `isHeaderNameValid()` function?

bretambrose: Could change it to validateHeaderName() I suppose.

source/event_stream.c
```diff
@@ -255,25 +260,607 @@ static void s_napi_event_stream_connection_on_connection_shutdown(
...
+    }
+
+    storage->message_type = message->message_type;
+    storage->message_flags = message->message_flags;
+
+    return AWS_OP_SUCCESS;
+
+error:
+
+    /* assumes AWS_ZERO_STRUCT was called first */
```
sbSteveK, line 316: It looks like 'aws_mem_calloc()' on the event that this storage is contained in might be zeroing out all relevant bits but I'm not seeing AWS_ZERO_STRUCT being explicitly called on the storage anywhere prior to this point. My guess is 'aws_mem_calloc()` covers this base but just calling it out in case AWS_ZERO_STRUCT should have been called on the storage specifically.

sbSteveK (approved):

source/event_stream.c
bretambrose: well zeroed memory is zeroed memory

PR #417, opened 2023-03-03
Event stream stream

bretambrose:
Stream bindings and tests

### DIFF
--- a/source/event_stream.c
+++ b/source/event_stream.c
@@ -1707,9 +1707,28 @@ static void s_napi_event_stream_on_stream_ended(napi_env env, napi_value functio
     struct aws_event_stream_client_stream_binding *binding = user_data;
 
     if (env && !binding->is_closed) {
+        napi_value params[1];
+        const size_t num_params = AWS_ARRAY_SIZE(params);
+
+        /*
+         * If we can't resolve the weak ref to the event stream, then it's been garbage collected and we
+         * should not do anything.
+         */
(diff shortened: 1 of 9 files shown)


source/event_stream.c
```diff
@@ -18,6 +18,23 @@ static const char *AWS_EVENT_STREAM_PROPERTY_NAME_HEADERS = "headers";
 static const char *AWS_EVENT_STREAM_PROPERTY_NAME_PAYLOAD = "payload";
 static const char *AWS_EVENT_STREAM_PROPERTY_NAME_FLAGS = "flags";
 static const char *AWS_EVENT_STREAM_PROPERTY_NAME_MESSAGE = "message";
+static const char *AWS_EVENT_STREAM_PROPERTY_NAME_OPERATION = "operation";
+
+/*
+ * The fact that a zeroed array list crashes in aws_array_list_length drives me crazy.  Since CBMC proofs are
```
sbSteveK, line 24: ooc, would it be worthwhile to add these additional checks to a separate function in common that wouldn't crash for a zeroed array list? Or is that unnecessary clutter and better left to being checked in this way wherever we may have a zeroed array list?

sbSteveK (approved):

source/event_stream.c
bretambrose: Unsure for now.  I would prefer to avoid touching array list in common due to the CBMC proofs and invariants.

PR #418, opened 2023-03-03
Eventstream tests in CI

bretambrose:
* Adds support for running the v2 Java SDK's EchoTest eventstream server in CI
* Misc fixes

### DIFF
--- a/.builder/actions/crt-ci-test.py
+++ b/.builder/actions/crt-ci-test.py
@@ -60,15 +60,14 @@ class CrtCiTest(Builder.Action):
             finally:
                 env.shell.popd()
 
-        return (proc, java_sdk_dir)
+        return proc, java_sdk_dir
 
     def run(self, env):
 
         # Unfortunately, we can't use NamedTemporaryFile and a with-block because NamedTemporaryFile is not readable
         # on Windows.
         try:
-
-            (proc, java_sdk_dir) = self._build_and_run_eventstream_echo_server(env)
+            proc, java_sdk_dir = self._build_and_run_eventstream_echo_server(env)
 
             env.shell.setenv("AWS_TESTING_COGNITO_IDENTITY", env.shell.get_secret("aws-c-auth-testing/cognito-identity"), quiet=True)
 
--- a/.builder/actions/crt-ci-test.py
+++ b/.builder/actions/crt-ci-test.py
@@ -27,14 +27,54 @@ class CrtCiTest(Builder.Action):
 
         return filename
 
-    def run(self, env):
-        env.shell.setenv("AWS_TESTING_COGNITO_IDENTITY", env.shell.get_secret("aws-c-auth-testing/cognito-identity"), quiet=True)
+    def _build_and_run_eventstream_echo_server(self, env):
+        java_sdk_dir = None
+        proc = None
 
-        self._write_environment_script_secret_to_env(env, "mqtt5-testing/github-ci-environment")
+        try:
+            env.shell.exec(["mvn", "--version"])
+
+            # maven is installed, so this is a configuration we can start an event stream echo server
+            java_sdk_dir = env.shell.mktemp()
+
+            env.shell.exec(["git", "clone", "https://github.com/aws/aws-iot-device-sdk-java-v2"], working_dir=java_sdk_dir)
+
+            sdk_dir = os.path.join(java_sdk_dir, "aws-iot-device-sdk-java-v2", "sdk")
+            env.shell.pushd(sdk_dir)
+
+            try:
+                env.shell.exec(["mvn", "compile"])
+                env.shell.exec(["mvn", "test", "-DskipTests=true"])
+                env.shell.exec(["mvn", "dependency:build-classpath", "-Dmdep.outputFile=classpath.txt"])
+
+                with open('classpath.txt', 'r') as file:
+                    classpath = file.read()
+
+                echo_server_command = ["java", "-classpath", f"{sdk_dir}/target/test-classes:$CURRENT_DIR/target/classes:{classpath}", "software.amazon.awssdk.eventstreamrpc.echotest.EchoTestServiceRunner", "127.0.0.1", "8033"]
+
+                # bypass builder's exec wrapper since it doesn't allow for background execution
+                proc = subprocess.Popen(echo_server_command)
+
+                env.shell.setenv("AWS_TEST_EVENT_STREAM_ECHO_SERVER_HOST", "127.0.0.1", quiet=False)
+                env.shell.setenv("AWS_TEST_EVENT_STREAM_ECHO_SERVER_PORT", "8033", quiet=False)
+            finally:
+                env.shell.popd()
+
+        return (proc, java_sdk_dir)
+
+    def run(self, env):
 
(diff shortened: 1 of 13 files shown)


.builder/actions/crt-ci-test.py
```diff
@@ -27,29 +27,89 @@ def _write_secret_to_temp_file(self, env, secret_name):
...
+            # maven is installed, so this is a configuration we can start an event stream echo server
+            java_sdk_dir = env.shell.mktemp()
+
+            env.shell.exec(["git", "clone", "https://github.com/aws/aws-iot-device-sdk-java-v2"], working_dir=java_sdk_dir, check=True)
+
+            sdk_dir = os.path.join(java_sdk_dir, "aws-iot-device-sdk-java-v2", "sdk")
+            env.shell.pushd(sdk_dir)
+
+            try:
+                env.shell.exec(["mvn", "compile"], check=True)
```
graebm, line 46: trivial: don't need to run both `mvn compile` and `mvn test`, just running `test` will implicitly run `compile`

> build phases ... are executed in the order given up to the point of the one specified.
https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#default-lifecycle


```suggestion
```

.builder/actions/crt-ci-test.py
```diff
@@ -27,29 +27,89 @@ def _write_secret_to_temp_file(self, env, secret_name):
...
+                env.shell.exec(["mvn", "test", "-DskipTests=true"], check=True)
```
graebm, line 47: This will probably do the trick, and makes a bit more sense than "test, but don't test"
```suggestion
                # the echo server is a test class
                env.shell.exec(["mvn", "test-compile"], check=True)
```

.builder/actions/crt-ci-test.py
```diff
@@ -27,29 +27,89 @@ def _write_secret_to_temp_file(self, env, secret_name):
...
+                test_class_path = os.path.join(sdk_dir, "target", "test-classes")
+                target_class_path = os.path.join(sdk_dir, "target", "classes")
+                directory_separator = os.pathsep
+
+                echo_server_command = ["java", "-classpath", f"{test_class_path}{directory_separator}{target_class_path}{directory_separator}{classpath}", "software.amazon.awssdk.eventstreamrpc.echotest.EchoTestServiceRunner", "127.0.0.1", "8033"]
+
+                print(f'Echo server command: {echo_server_command}')
+
+                # bypass builder's exec wrapper since it doesn't allow for background execution
+                proc = subprocess.Popen(echo_server_command)
```
graebm, line 62: Check out how Dengke set up the background process to get killed when builder exited:
https://github.com/awslabs/aws-c-http/blob/270f0968fd6b5317b8dd714b2bbc6fd04976dbcf/.builder/action/local-server-setup.py#L43

.builder/actions/crt-ci-test.py
```diff
@@ -27,29 +27,89 @@ def _write_secret_to_temp_file(self, env, secret_name):
...
+
+                env.shell.setenv("AWS_TEST_EVENT_STREAM_ECHO_SERVER_HOST", "127.0.0.1", quiet=False)
+                env.shell.setenv("AWS_TEST_EVENT_STREAM_ECHO_SERVER_PORT", "8033", quiet=False)
+            except:
+                raise
```
graebm, line 67: trivial: you can just omit this, since you have a finally block
```suggestion
```

graebm (commented):
trivial comments, but overall pretty darn cool

source/event_stream.c
```diff
@@ -1541,36 +1569,875 @@ napi_value aws_napi_event_stream_client_connection_send_protocol_message(napi_en
...
-    napi_throw_error(env, NULL, "aws_napi_event_stream_client_stream_new - NYI");
+    /*
+     * We ref count the binding itself because there are two primary time intervals that together create a union
+     * that we must honor.
+     *
+     * Interval #1: The binding must live from new() to extern finalizer, which is only triggered by a call to close()
+     * Interval #2: The binding must live from activate() to {stream failure || stream shutdown} as processed
+     *    by the libuv thread.  It is incorrect to react to those events in the event loop callback; we must bundle
+     *    and ship them across to the libuv thread.  When the libuv thread is processing a stream failure or
+     *    shutdown, we know that no other events can possibly be pending ()hey would have already been
```
TwistedTwigleg, line 1596: Extremely minor: `... pending ()hey would...` -> `... pending they would ....`

source/event_stream.c
```diff
@@ -1541,36 +1569,875 @@ napi_value aws_napi_event_stream_client_connection_send_protocol_message(napi_en
...
+    napi_ref node_event_stream_client_stream_ref = binding->node_event_stream_client_stream_ref;
+    binding->node_event_stream_client_stream_ref = NULL;
+
+    if (node_event_stream_client_stream_external_ref != NULL) {
+        napi_delete_reference(env, node_event_stream_client_stream_external_ref);
+    }
+
+    if (node_event_stream_client_stream_ref != NULL) {
+        napi_delete_reference(env, node_event_stream_client_stream_ref);
+    }
```
TwistedTwigleg, line 1678: Question: Could this be simplified to the following?

~~~
if (binding->node_event_stream_client_stream_ref != NULL) {
    napi_delete_reference(env, binding->node_event_stream_client_stream_ref);
    binding->node_event_stream_client_stream_ref = NULL
}
if (binding->node_event_stream_client_stream_ref != NULL) {
    napi_delete_reference(env, binding->node_event_stream_client_stream_ref);
    binding->node_event_stream_client_stream_ref = NULL
}
~~~

Not that it is bad by any means how it is currently, I'm just curious if there is a reason for using a couple local variables (maybe threading issues?). I am guessing there is probably something I'm missing and so I wanted to ask.

source/event_stream.c
```diff
@@ -1541,36 +1569,875 @@ napi_value aws_napi_event_stream_client_connection_send_protocol_message(napi_en
...
+    if (s_aws_extract_activation_options_from_js(
+            env, napi_activation_options, binding, &operation_name_buffer, &activation_message)) {
+        napi_throw_error(
+            env,
+            NULL,
+            "aws_napi_event_stream_client_stream_activate - unable to unpack activation options from JS object");
+        goto done;
+    }
+
+    napi_value stream_activation_callback = *arg++;
```
TwistedTwigleg, line 2219: Question: Does this need to be checked for null?

TwistedTwigleg (approved):
Looks good to me!

source/event_stream.c
bretambrose: Assignment to null is intentionally done first.  It may not be necessary here, but in general, you'll see similar things in aws-c-* when we erase something before possibly triggering external callbacks that could reach back in.

source/event_stream.c
bretambrose: It's synthesized internally, so no.

That being said, I'm going to do an invalid input pass after this with tests and probably some fixes where the tests uncover unhelpful failures.

source/event_stream.c
TwistedTwigleg: Ah okay, that makes sense and is good to know! I figured there had to be a reason but was unsure.

PR #421, opened 2023-03-09
Experimental: Babel pre-process step

bretambrose:
cmake-js is an irreplacable dependency of the CRT.  Starting with major version 7.x, cmake-js uses language features only found in node14+.  In order to continue to be able to upgrade cmake-js without having to modify our node baseline to 14, this PR adds a babel-driven pre-processing step that rewrites the options-chaining code in cmake-js to be usable from earlier versions of node.


xiazhvera (approved):

package.json
```diff
@@ -39,22 +39,24 @@
...
+    "yargs": "^17.2.1"
   },
   "dependencies": {
     "@aws-sdk/util-utf8-browser": "^3.109.0",
+    "@babel/cli": "^7.21.0",
+    "@babel/core": "^7.21.0",
+    "@babel/plugin-proposal-optional-chaining": "^7.21.0",
     "@httptoolkit/websocket-stream": "^6.0.0",
     "axios": "^0.24.0",
+    "cmake-js": "7.1.1",
```
TingDaoK, line 57: why do we pin to `7.1.1` instead of using `^7.1.1`?

bretambrose: 7.2.x bumps axios to a version that causes import errors on x86 node v12

TingDaoK (approved):

graebm:
Closing this. We're upgrading to Node 14 soon to avoid hacky workarounds

PR #424, opened 2023-03-10
Event stream cleanup assist

bretambrose:
Experimental PR to help mitigate potential leaks from not following the eventstream API contract.

* Adds automatic close() invocations to eventstream connection and stream when native notifies of disconnect/termination.
* Adds automatic terminate stream message when closing an open stream.


bretambrose:
On second thought, this isn't the right place to do auto-cleanup.  Event ordering issues make the RPC client a much better choice where multiple promises can be unioned before invoking close.

PR #430, opened 2023-03-15
Downgrade package-lock to version 1 

xiazhvera:
*Issue #, if available:*

*Description of changes:*
Downgrade package-lock to version 1 to make sure support node6

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6891 +1,8 @@
 {
   "name": "aws-crt",
   "version": "1.0.0-dev",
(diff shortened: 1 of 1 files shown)


package-lock.json
graebm, line 1: If we need this to stay at version 1, then we need an explicit CI check to make sure we don't accidentally change it by accident

it gets accidentally changed all the time when people upgrade packages on their local machines

xiazhvera:
Need further discussion on the selected version. Close it for now.

PR #431, opened 2023-03-16
Improve browser MQTT interaction with credentials refresh

bretambrose:
* Extend the browser credentials provider with an async refresh method
* Update MQTT311 and MQTT5 clients to use the async refresh
* Update browser mqtt sample to use new method and refactor to use a simpler credentials provider wrapper


lib/browser/auth.ts
```diff
@@ -48,6 +48,12 @@ export class CredentialsProvider {
     { 
         return undefined;
     }
+
+    async refreshCredentials() : Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            resolve();
```
TwistedTwigleg, line 54: Question: Shouldn't this call `getCredentials`? Doesn't the code, as written, just automatically resolve...?

TwistedTwigleg: Or is this function extended somewhere? I didn't see it in this PR nor when I looked in the code, but maybe I missed it?

TwistedTwigleg: Ah, never mind, I see it now in the sample. This function is supposed to be overridden and so, by default, it just returns true. I'd add a comment on this function to explain what it's for and that it needs to be overridden/extended to return a promise that resolves when the customer's code has refreshed the credentials.

TwistedTwigleg (approved):

PR #432, opened 2023-03-16
X509 binding

TwistedTwigleg:
*Description of changes:*

Binds X509 support to the Python AWS credentials provider and adds the tests to make sure it works as expected (modeled tests as close to those in Java).

Only adds X509 support to NodeJS - the browser version was not modified.

________


lib/native/auth.ts
```diff
@@ -80,6 +80,47 @@ export interface CognitoCredentialsProviderConfig {
...
+     * X509 service regional endpoint to source credentials from.
+     * This is a per-account value that can be determined via the CLI:
+     * `aws iot describe-endpoint --endpoint-type iot:CredentialProvider`
+     */
+    endpoint: string;
+
+    /**
+     * The name of the IoT thing to use to fetch credentials.
+     */
+    thing_name: string;
```
bretambrose, line 100: camel case for consistency here.  We're not consistent across the library, but I think camel case should be what we stick with going forward.

lib/native/auth.ts
```diff
@@ -80,6 +80,47 @@ export interface CognitoCredentialsProviderConfig {
...
+    /**
+     * The name of the role alias to fetch credentials through.
+     */
+    role_alias: string;
+
+    /**
+     * TLS context for secure socket connections.
+     * If undefined, then a default tls context will be created and used.
+     */
+    tlsContext?: ClientTlsContext;
```
bretambrose, line 111: The credentials provider will not work without a mTLS-derived tls context.  This should be mandatory.

lib/native/auth.ts
```diff
@@ -139,6 +180,23 @@ export class AwsCredentialsProvider extends crt_native.AwsCredentialsProvider {
...
+     * @param config provider configuration necessary to source credentials via X509
+     *
+     * @returns a new credentials provider that returns credentials sourced from the AWS X509 service
+     */
+    static newX509(config : X509CredentialsConfig): AwsCredentialsProvider {
+        if (config == null || config == undefined) {
+            throw new CrtError("AwsCredentialsProvider newX509: X509 config not defined")
+        }
+        return super.newX509(config,
+            config.tlsContext != null ? config.tlsContext.native_handle() : new ClientTlsContext().native_handle(),
```
bretambrose, line 196: remove the create-if-null based on above comment

lib/native/auth.ts
```diff
@@ -80,6 +80,47 @@ export interface CognitoCredentialsProviderConfig {
...
+    /**
+     * TLS context for secure socket connections.
+     * If undefined, then a default tls context will be created and used.
+     */
+    tlsContext?: ClientTlsContext;
+
+    /**
+     * Client bootstrap to use.  In almost all cases, this can be left undefined.
+     */
+    bootstrap?: ClientBootstrap;
```
bretambrose, line 116: I wouldn't even bother with this and just use the default inside the native code.

bretambrose (approved):

TwistedTwigleg:
Thanks! Enabling auto-merge so it can be merged into `main`...

PR #447, opened 2023-04-21
Add token_key_name and token_value support to MQTT311 custom auth

TwistedTwigleg:
*Description of changes:*

The current MQTT311 custom auth implementation is missing the ability to set the custom authorizer token key name and token key value, making it impossible to login with a signed custom authorizer using the builder.

This PR adds the support for the custom auth token key name and token key value. The change is completely backwards compatible as the new values are optional.

(Though the V2 SDKs will need to be updated because at least the browser refers to the custom auth signature as the "password" which is incorrect...)

________

### DIFF
--- a/lib/common/aws_iot_shared.ts
+++ b/lib/common/aws_iot_shared.ts
@@ -58,7 +58,8 @@ import * as mqtt5_packet from "./mqtt5_packet";
  */
 export function populate_username_string_with_custom_authorizer(
     current_username? : string, input_username? : string, input_authorizer? : string,
-    input_signature? : string, input_builder_username? : string) {
+    input_signature? : string, input_builder_username? : string,
+    input_token_key_name? : string, input_token_value? : string) {
 
     let username_string = "";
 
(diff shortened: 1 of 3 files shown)


bretambrose (commented):
Can we add a full signed authorizer test for 311 that just mirrors what we have for 5?

TwistedTwigleg:
Sure, I'll  add a test and will update C++ (and later Java) as well.

lib/browser/aws_iot.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+test_utils.conditional_test(test_utils.ClientEnvironmentalConfig.hasIotCoreEnvironment())('Aws Iot Core Mqtt over websockets with Non-Signing Custom Auth - Connection Success', async () => {
+
+    let builder = aws_iot_mqtt311.AwsIotMqttConnectionConfigBuilder.new_default_builder();
+    builder.with_custom_authorizer(
+        test_utils.ClientEnvironmentalConfig.AWS_IOT_NO_SIGNING_AUTHORIZER_USERNAME,
+        test_utils.ClientEnvironmentalConfig.AWS_IOT_NO_SIGNING_AUTHORIZER_NAME,
+        "",
+        test_utils.ClientEnvironmentalConfig.AWS_IOT_NO_SIGNING_AUTHORIZER_PASSWORD,
+        undefined,
```
bretambrose, line 23: No need to explicitly say undefined here

lib/common/aws_iot_shared.ts
```diff
@@ -52,13 +52,16 @@ import * as mqtt5_packet from "./mqtt5_packet";
  * @param input_authorizer the name of the authorizer to add - can be an empty string to skip
  * @param input_signature the name of the signature to add - can be an empty string to skip
  * @param input_builder_username the username from the MQTT builder
+ * @param input_token_key_name the token key name - can be empty string to skip
```
bretambrose, line 55: empty string comment is kinda weird; these are optional parameters.

bretambrose (approved):

TwistedTwigleg:

PR #451, opened 2023-05-08
Let's try ref counting

bretambrose:
Revert the changes that prevented multi-init of the NAPI module and add ref-counting support and careful, matching shutdown when the last ref goes away.

https://github.com/aws/aws-iot-device-sdk-js-v2/discussions/360

A more complex solution might consider making elg/hr/bootstrap a per-environment set of data, but for now, the simplest approach with the least possibility of error is to simply have all environments reference a shared triple and use ref-counting to drop the triple once no env-users remain.

### DIFF
--- a/source/module.c
+++ b/source/module.c
@@ -1074,31 +1074,66 @@ static void s_install_crash_handler(void) {
 #endif
 }
 
+static void s_uninstall_crash_handler(void) {
+#if defined(_WIN32)
+    SetUnhandledExceptionFilter(NULL);
+#elif defined(AWS_HAVE_EXECINFO)
+    sigaction(SIGSEGV, NULL, NULL);
+    sigaction(SIGABRT, NULL, NULL);
+    sigaction(SIGILL, NULL, NULL);
+    sigaction(SIGBUS, NULL, NULL);
+#endif
+}
+
+/*
+ * Temporary hack to detect multi-init so we can throw an exception because we haven't figured out the right way
+ * to support it yet.  Better than a hard crash in native code.
+ */
+static struct aws_mutex s_module_lock = AWS_MUTEX_INIT;
+static uint32_t s_module_initialize_count = 0;
+
 static void s_napi_context_finalize(napi_env env, void *user_data, void *finalize_hint) {
     (void)env;
     (void)finalize_hint;
-    aws_client_bootstrap_release(s_default_client_bootstrap);
-    aws_host_resolver_release(s_default_host_resolver);
-    aws_event_loop_group_release(s_node_uv_elg);
 
-    aws_thread_join_all_managed();
+    aws_mutex_lock(&s_module_lock);
+    --s_module_initialize_count;
 
-    aws_unregister_log_subject_info_list(&s_log_subject_list);
-    aws_unregister_error_info(&s_error_list);
-    aws_auth_library_clean_up();
-    aws_mqtt_library_clean_up();
+    if (s_module_initialize_count == 0) {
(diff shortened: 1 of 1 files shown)


source/module.c
```diff
@@ -1136,76 +1171,65 @@ static bool s_create_and_register_function(
...
-    AWS_FATAL_ASSERT(s_default_client_bootstrap == NULL);
+    if (s_module_initialize_count == 0) {
+        aws_cal_library_init(allocator);
+        aws_http_library_init(allocator);
+        aws_mqtt_library_init(allocator);
+        aws_auth_library_init(allocator);
+        aws_register_error_info(&s_error_list);
+        aws_register_log_subject_info_list(&s_log_subject_list);
+
+        s_install_crash_handler();
```
graebm, line 1188: trivial: maybe install the crash handler earlier? (for diagnosing any weirdness from a library_init() call)

source/module.c
```diff
@@ -1074,31 +1074,66 @@ static void s_install_crash_handler(void) {
...
+        aws_event_loop_group_release(s_node_uv_elg);
+        s_node_uv_elg = NULL;
 
-    aws_unregister_log_subject_info_list(&s_log_subject_list);
-    aws_unregister_error_info(&s_error_list);
-    aws_auth_library_clean_up();
-    aws_mqtt_library_clean_up();
+        aws_thread_join_all_managed();
+
+        s_uninstall_crash_handler();
```
graebm, line 1114: trivial: maybe uninstall the crash handler later on?

source/module.c
```diff
@@ -1074,31 +1074,66 @@ static void s_install_crash_handler(void) {
...
 static void s_napi_context_finalize(napi_env env, void *user_data, void *finalize_hint) {
     (void)env;
     (void)finalize_hint;
-    aws_client_bootstrap_release(s_default_client_bootstrap);
-    aws_host_resolver_release(s_default_host_resolver);
-    aws_event_loop_group_release(s_node_uv_elg);
 
-    aws_thread_join_all_managed();
+    aws_mutex_lock(&s_module_lock);
+    --s_module_initialize_count;
```
graebm, line 1100: debatable:
```suggestion
    AWS_FATAL_ASSERT(s_module_initialize_count > 0);
    --s_module_initialize_count;
```

graebm (approved):
Nice work!!!
I'd like to do a pass on all our library_init() calls to make them work like this someday, protected by a lock and refcount.

source/module.c
```diff
@@ -1074,31 +1074,66 @@ static void s_install_crash_handler(void) {
...
+    sigaction(SIGABRT, NULL, NULL);
+    sigaction(SIGILL, NULL, NULL);
+    sigaction(SIGBUS, NULL, NULL);
+#endif
+}
+
+/*
+ * Temporary hack to detect multi-init so we can throw an exception because we haven't figured out the right way
+ * to support it yet.  Better than a hard crash in native code.
+ */
```
TingDaoK, line 1091: this comment looks out of date, we are handling it now

TingDaoK (approved):

source/module.c
bretambrose: > trivial: maybe uninstall the crash handler later on?

Originally I was worried that the backtrace usage would require init to have happened first (for example, if we called aws_backgrace_log) but since we use aws_backtrace_print and the impl is only hitting posix functions and one byte buffer usage, I think it's safe to make the crash handler wrap the library init/cleanup.

PR #455, opened 2023-05-19
Adjust MQTT311 disconnect cleanup

TwistedTwigleg:
*Description of changes:*

~Adjusts MQTT311's disconnect behavior to not always free the native resources when `disconnect()` is called, instead adding a new option (`close_on_disconnect`) that can be set to `false`, allowing for manual reconnects. Also exposes the `close()` function so the connection can be cleaned manually if `close_on_disconnect` is set to `false`.~

~Also fixes a number of segfaults that can occur if the MQTT311 connection is closed. Now errors are thrown explaining that the MQTT311 connection has been closed and is no longer usable, rather than segfaulting.~

~Adding documentation for these new functions and tests to ensure functionality works as expected by default, and with the new option.~

Fixes segfaults that occur when you call `connect`, `disconnect`, and then `connect` again, or if you call `disconnect` and then try to perform any other operations (subscribe, unsubscribe, publish, etc). Also adjusts documentation a bit, and adds a couple tests to ensure the current behavior works as expected without segfaults.

_________

### DIFF
--- a/lib/native/aws_iot.ts
+++ b/lib/native/aws_iot.ts
@@ -73,8 +73,7 @@ export class AwsIotMqttConnectionConfigBuilder {
             password: undefined,
             tls_ctx: undefined,
             reconnect_min_sec: DEFAULT_RECONNECT_MIN_SEC,
-            reconnect_max_sec: DEFAULT_RECONNECT_MAX_SEC,
-            close_on_disconnect: true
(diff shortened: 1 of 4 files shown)


xiazhvera (approved):

TwistedTwigleg:

PR #476, opened 2023-07-05
Update For V8 Memory Cage

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Add cmake flag to disable napi external buffer support for memory cage. [Electron and V8 memory cage](https://www.electronjs.org/blog/v8-memory-cage)
- Disable binding callbacks on client close

### DIFF
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -148,9 +148,9 @@ static void s_on_body_call(napi_env env, napi_value on_body, void *context, void
 
 #else
         void *buf_data = NULL;
-        AWS_NAPI_ENSURE(env, napi_create_arraybuffer(env, args->chunk.len, buf_data, &params[0]));
+        AWS_NAPI_ENSURE(env, napi_create_arraybuffer(env, args->chunk.len, &buf_data, &params[0]));
 
-        memcpy(args->chunk.buffer, buf_data, args->chunk.len);
+        memcpy(buf_data, args->chunk.buffer, args->chunk.len);
 
         // As the chunk is copied into NodeJS, release the data
         aws_byte_buf_clean_up(&args->chunk);
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -138,8 +138,8 @@ static void s_on_body_call(napi_env env, napi_value on_body, void *context, void
         napi_value params[1];
         const size_t num_params = AWS_ARRAY_SIZE(params);
 
-        aws_napi_create_external_arraybuffer_function(
-            env, args->chunk.buffer, args->chunk.len, s_external_arraybuffer_finalizer, args, &params[0]);
+        AWS_NAPI_ENSURE(aws_napi_create_external_arraybuffer(
+            env, args->chunk.buffer, args->chunk.len, s_external_arraybuffer_finalizer, args, &params[0]));
 
         AWS_NAPI_ENSURE(
             env, aws_napi_dispatch_threadsafe_function(env, binding->on_body, NULL, on_body, num_params, params));
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -138,8 +138,10 @@ static void s_on_body_call(napi_env env, napi_value on_body, void *context, void
         napi_value params[1];
         const size_t num_params = AWS_ARRAY_SIZE(params);
 
-        AWS_NAPI_ENSURE(aws_napi_create_external_arraybuffer(
-            env, args->chunk.buffer, args->chunk.len, s_external_arraybuffer_finalizer, args, &params[0]));
+        AWS_NAPI_ENSURE(
+            env,
+            aws_napi_create_external_arraybuffer(
+                env, args->chunk.buffer, args->chunk.len, s_external_arraybuffer_finalizer, args, &params[0]));
 
         AWS_NAPI_ENSURE(
             env, aws_napi_dispatch_threadsafe_function(env, binding->on_body, NULL, on_body, num_params, params));
--- a/source/http_stream.c
+++ b/source/http_stream.c
@@ -119,6 +119,7 @@ struct on_body_args {
     struct aws_allocator *allocator;
 };
 
+#ifndef NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED
 static void s_external_arraybuffer_finalizer(napi_env env, void *finalize_data, void *finalize_hint) {
     (void)env;
     (void)finalize_data;
@@ -126,6 +127,7 @@ static void s_external_arraybuffer_finalizer(napi_env env, void *finalize_data, 
     aws_byte_buf_clean_up(&args->chunk);
     aws_mem_release(args->allocator, args);
 }
+#endif
 
 static void s_on_body_call(napi_env env, napi_value on_body, void *context, void *user_data) {
     struct http_stream_binding *binding = context;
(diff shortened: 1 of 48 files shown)


source/http_stream.c
```diff
@@ -138,11 +140,23 @@ static void s_on_body_call(napi_env env, napi_value on_body, void *context, void
         napi_value params[1];
         const size_t num_params = AWS_ARRAY_SIZE(params);
 
+#ifndef NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED
```
graebm, line 143: We could do a runtime check instead. The [Electron docs](https://www.electronjs.org/blog/v8-memory-cage#i-want-to-refactor-a-node-native-module-to-support-electron-21-how-do-i-do-that) say that `napi_create_external_buffer()` will fail. We could just try making the buffer the OTHER way if the first try fails.

The advantage is: the same compiled code will work in Electron/Non-Electron

Just an idea...

xiazhvera: lol Bret just made the same comment on slack. I will update it to do a runtime check instead.

source/module.c
```diff
@@ -841,6 +835,13 @@ const char *aws_napi_status_to_str(napi_status status) {
         case napi_bigint_expected:
             reason = "napi_bigint_expected";
             break;
+/* TODO: The enum `napi_no_external_buffers_allowed` is introduced in node14.
```
bretambrose, line 838: Is there a NAPI_VERSION value we could use here instead of commenting it out?

bretambrose (approved):

source/module.c
```diff
@@ -268,15 +268,9 @@ int aws_napi_attach_object_property_binary_as_finalizable_external(
...
-        napi_create_external_arraybuffer(
-            env,
-            data_buffer->buffer,
-            data_buffer->len,
-            s_finalize_external_binary_byte_buf,
-            data_buffer,
-            &napi_binary));
+
+    aws_napi_create_external_arraybuffer_function(
+        env, data_buffer->buffer, data_buffer->len, s_finalize_external_binary_byte_buf, data_buffer, &napi_binary);
```
graebm, line 273: this isn't checking the error anymore

graebm (changes requested):

source/module.c
```diff
@@ -931,6 +932,37 @@ static void s_handle_failed_callback(napi_env env, napi_value function, napi_sta
...
+        AWS_NAPI_ENSURE(env, napi_create_arraybuffer(env, byte_length, &napi_buf_data, result));
+
+        memcpy(napi_buf_data, external_data, byte_length);
+
+        // As the data has been copied into the Node, invoke the finalize callback to make sure the
+        // data is released.
+        finalize_cb(env, finalize_hint, finalize_hint);
+    }
+
+    return external_buffer_status;
```
graebm, line 963: if we take the if-branch, and then succeed, `external_buffer_status` will still have an error value

source/module.c
xiazhvera: Looking into the napi header files, I didn't see the NAPI_VERSION used for the enum. Meanwhile, I also didnt see any NAPI_VERSION ties to `napi_queue_full`, however, we did use NAPI_VERSION to opt in the enum value. I'm curious how could we determin which NAPI_VERSION the enum is introduced?

source/module.c
xiazhvera: Updated the `aws_napi_create_external_arraybuffer`. Now it would return the error code from `napi_create_arraybuffer` instead of assert the function when failed. It will also return `napi_ok` for a successful path. The caller would be responsible to handle the `exteranl_buffer` when the function failed.

source/module.c
```diff
@@ -931,6 +939,48 @@ static void s_handle_failed_callback(napi_env env, napi_value function, napi_sta
...
+
+    if (external_buffer_status != napi_ok) {
+
+        /* TODO: The enum `napi_no_external_buffers_allowed` is introduced in node14.
+         * Use it to determine if the function failed because of the external buffer support after bump
+         * minimal support to node 14
+         */
+        AWS_NAPI_LOGF_ERROR(
+            "napi_create_external_arraybuffer (in aws_napi_create_external_arraybuffer) failed with : %s",
+            aws_napi_status_to_str(external_buffer_status));
```
graebm, line 961: In the current code, this is going to log at the ERROR level, even though the next `napi_create_arraybuffer()` call is likely to succeed

Customers get **very scared** by ERROR logs

Don't log unless both attempts to make the buffer fail.

source/mqtt5_client.c
```diff
@@ -3517,6 +3527,8 @@ napi_value aws_napi_mqtt5_client_close(napi_env env, napi_callback_info info) {
         return NULL;
     }
 
+    s_aws_mqtt5_release_callbacks(binding);
```
bretambrose, line 3530: This is not safe to do here.  We need to dig a lot deeper to find out what's going on here to cause problems.

bretambrose (changes requested):

xiazhvera:
Need further investigation.

xiazhvera:
Removed some of the electron support changes. The PR is now for V8 Memory Cage only.

bretambrose (approved):

source/module.c
```diff
@@ -841,6 +848,10 @@ const char *aws_napi_status_to_str(napi_status status) {
         case napi_bigint_expected:
```
graebm, line 848: A few lines up, there's a `#if NAPI_VERSION >= 3`.

I got stressed out about how the new `case` here is within a conditional.

Then I did some research and realize this file [won't even compile](https://github.com/awslabs/aws-crt-nodejs/blob/96f3e3425d6d07024726ac19d8b14d453a7bb0cc/source/module.c#L46) unless NAPI_VERSION is 4 or higher.

So please delete the `#if NAPI_VERSION >= 3`, it serves no purpose anymore

source/module.c
```diff
@@ -931,6 +942,46 @@ static void s_handle_failed_callback(napi_env env, napi_value function, napi_sta
...
+
+    napi_status external_buffer_status =
+        napi_create_external_arraybuffer(env, external_data, byte_length, finalize_cb, finalize_hint, result);
+
+    if (external_buffer_status == NAPI_NO_EXTERNAL_BUFFER_ENUM_VALUE) {
+
+        AWS_LOGF_DEBUG(
+            AWS_LS_NODEJS_CRT_GENERAL,
+            "napi_create_external_arraybuffer (in aws_napi_create_external_arraybuffer) failed with : %s",
+            aws_napi_status_to_str(external_buffer_status));
```
graebm, line 961: don't log this.
If a node installation doesn't support external buffers, this is going to print for every single allocation. It's just too much. Especially at DEBUG level.

source/module.c
```diff
@@ -931,6 +942,46 @@ static void s_handle_failed_callback(napi_env env, napi_value function, napi_sta
...
+
+        // The external buffer is disabled, manually copy the external_data into Node
+        void *napi_buf_data = NULL;
+        napi_status create_arraybuffer_status = napi_create_arraybuffer(env, byte_length, &napi_buf_data, result);
+
+        if (create_arraybuffer_status != napi_ok) {
+            AWS_LOGF_DEBUG(
```
graebm, line 968: AWS_LOGF_ERROR?
I mean, failing to allocate the buffer is a pretty big deal, right?

source/module.c
```diff
@@ -931,6 +939,40 @@ static void s_handle_failed_callback(napi_env env, napi_value function, napi_sta
...
+                "napi_create_arraybuffer (in aws_napi_create_external_arraybuffer) failed with : %s",
+                aws_napi_status_to_str(create_arraybuffer_status));
+            return create_arraybuffer_status;
+        }
+
+        memcpy(napi_buf_data, external_data, byte_length);
+
+        // As the data has been copied into the Node, invoke the finalize callback to make sure the
+        // data is released.
+        finalize_cb(env, finalize_hint, finalize_hint);
```
xiazhvera, line 970: Debatable: Currently I do not invoke `finalize_cb` if the creation failed. The function could also call the `finalize_cb` when the arraybuffer creation failed to avoid memory leak. Not sure which way is a better though.

source/module.c
```diff
@@ -47,6 +47,12 @@ AWS_STATIC_ASSERT(NAPI_VERSION >= 4);
 
 #define AWS_DEFINE_ERROR_INFO_CRT_NODEJS(CODE, STR) AWS_DEFINE_ERROR_INFO(CODE, STR, "aws-crt-nodejs")
 
+/* TODO:
+ * Hardcoded enum value for `napi_no_external_buffers_allowed`.
+ * The enum `napi_no_external_buffers_allowed` is introduced in node14.
+ * Use it for external buffer related changes after bump to node 14 */
+#define NAPI_NO_EXTERNAL_BUFFER_ENUM_VALUE 22
```
graebm, line 54: we don't need this anymore. we have node 14, we can use the actual enum

source/module.c
graebm: this is correct. You shouldn't invoke the finalizer if you're reporting an error. It's up to whatever calls aws_napi_create_external_arraybuffer() to ensure memory's cleaned up in the case of failure.

graebm (approved):
fix & ship

PR #489, opened 2023-09-28
Add success and failure callbacks

alfred2g:
Add on success and on failure callbacks to the bindings and to mqtt.ts

### DIFF
--- a/lib/native/mqtt.ts
+++ b/lib/native/mqtt.ts
@@ -310,6 +310,8 @@ export class MqttClientConnection extends NativeResourceMixin(BufferedEventEmitt
             client.native_handle(),
             (error_code: number) => { this._on_connection_interrupted(error_code); },
             (return_code: number, session_present: boolean) => { this._on_connection_resumed(return_code, session_present); },
+            (return_code: number, session_present: boolean) => { this._on_connection_success(return_code, session_present); },
+            (error_code: number) => { this._on_connection_failure(error_code); },
             config.tls_ctx ? config.tls_ctx.native_handle() : null,
             will,
             config.username,
@@ -625,6 +627,18 @@ export class MqttClientConnection extends NativeResourceMixin(BufferedEventEmitt
                 this.emit('error', new CrtError(reason));
             });
         };
+    }
+
+    private _on_connection_failure(error_code: number) {
+        let failureCallbackData = { error: new CrtError(error_code) } as OnConnectionFailedResult;
+        this.emit('connection_failure', failureCallbackData);
+
+    }
+    private _on_connection_success(return_code: number, session_present: boolean) {
+
+        let successCallbackData = { session_present: session_present, reason_code: return_code } as OnConnectionSuccessResult;
+        this.emit('connection_success', successCallbackData);
+
     }
 
     private _on_connection_interrupted(error_code: number) {
--- a/source/mqtt_client_connection.c
+++ b/source/mqtt_client_connection.c
@@ -45,6 +45,8 @@ struct mqtt_connection_binding {
     napi_threadsafe_function on_any_publish;
     napi_threadsafe_function transform_websocket;
     napi_threadsafe_function on_closed;
+    napi_threadsafe_function on_connection_success;
+    napi_threadsafe_function on_connection_failure;
 };
 
 static void s_mqtt_client_connection_release_threadsafe_function(struct mqtt_connection_binding *binding) {
@@ -75,6 +77,18 @@ static void s_mqtt_client_connection_release_threadsafe_function(struct mqtt_con
     if (binding->on_closed != NULL) {
         AWS_NAPI_ENSURE(binding->env, aws_napi_release_threadsafe_function(binding->on_closed, napi_tsfn_abort));
         binding->on_closed = NULL;
+    }
+
+    if (binding->on_connection_success != NULL) {
+        AWS_NAPI_ENSURE(
+            binding->env, aws_napi_release_threadsafe_function(binding->on_connection_success, napi_tsfn_abort));
+        binding->on_connection_success = NULL;
+    }
+
+    if (binding->on_connection_failure != NULL) {
+        AWS_NAPI_ENSURE(
+            binding->env, aws_napi_release_threadsafe_function(binding->on_connection_failure, napi_tsfn_abort));
+        binding->on_connection_failure = NULL;
     }
 }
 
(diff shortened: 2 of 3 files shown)


lib/native/mqtt.ts
```diff
@@ -627,6 +629,18 @@ export class MqttClientConnection extends NativeResourceMixin(BufferedEventEmitt
...
+        let failureCallbackData = { error: new CrtError(error_code) } as OnConnectionFailedResult;
+        this.emit('connection_failure', failureCallbackData);
+
+    }
+    private _on_connection_success(return_code: number, session_present: boolean) {
+
+        let successCallbackData = { session_present: session_present, reason_code: return_code } as OnConnectionSuccessResult;
+        this.emit('connection_success', successCallbackData);
+
+    }
```
sfod, line 642: If I'm not mistaken, in some cases both `_on_connection_resumed` and `_on_connection_success` will be called together:
https://github.com/awslabs/aws-c-mqtt/blob/main/source/client_channel_handler.c#L242
https://github.com/awslabs/aws-c-mqtt/blob/main/source/client_channel_handler.c#L263

So, `this.emit('connection_success', successCallbackData);` should be removed from `_on_connection_resumed `. Probably other functions (e.g. `_on_connect_callback`) should be checked as well.

I couldn't find on_resume callback usage in tests (it's tricky to test it on this layer), but `_on_connect_callback` should be called on every successful connect. So, as an idea, you could add a check to the tests that `on_connection_success`/`on_connection_failure` are fired exactly once.

sfod (changes requested):

source/mqtt_client_connection.c
```diff
@@ -342,6 +486,11 @@ napi_value aws_napi_mqtt_client_connection_new(napi_env env, napi_callback_info
             binding->connection, s_on_connection_interrupted, binding, s_on_connection_resumed, binding);
     }
 
+    if (binding->on_connection_failure || binding->on_connection_success) {
+        aws_mqtt_client_connection_set_connection_result_handlers(
```
sfod, line 490: This call can fail, so we should check the return code

lib/native/mqtt.ts
```diff
@@ -627,6 +629,18 @@ export class MqttClientConnection extends NativeResourceMixin(BufferedEventEmitt
         };
     }
 
+    private _on_connection_failure(error_code: number) {
+        let failureCallbackData = { error: new CrtError(error_code) } as OnConnectionFailedResult;
+        this.emit('connection_failure', failureCallbackData);
+
+    }
+    private _on_connection_success(return_code: number, session_present: boolean) {
+
```
sfod, line 638: nit: Extra empty line

source/mqtt_client_connection.c
```diff
@@ -251,11 +265,113 @@ static void s_on_connection_resumed(
...
+    bool session_present,
+    void *user_data) {
+    (void)connection;
+
+    struct mqtt_connection_binding *binding = user_data;
+    if (!binding->on_connection_success) {
+        return;
+    }
+
+    struct connection_resumed_args *args =
```
sfod, line 313: Should be `struct connection_success_args`

lib/native/mqtt.ts
```diff
@@ -655,8 +665,6 @@ export class MqttClientConnection extends NativeResourceMixin(BufferedEventEmitt
         if (error_code == 0 && return_code == 0) {
             resolve(session_present);
             this.emit('connect', session_present);
-            let successCallbackData = { session_present: session_present, reason_code: return_code } as OnConnectionSuccessResult;
-            this.emit('connection_success', successCallbackData);
         } else if (error_code != 0) {
             reject("Failed to connect: " + io.error_code_to_string(error_code));
             let failureCallbackData = { error: new CrtError(error_code) } as OnConnectionFailedResult;
```
xiazhvera, line 670: We probably wanna remove `connection_failure` as well, as now it has its own callback.

source/mqtt_client_connection.c
```diff
@@ -87,6 +105,7 @@ static void s_mqtt_client_connection_finalize(napi_env env, void *finalize_data,
 
     /* Should have already been done, but just to be safe -- now that it's reentrant -- release the functions anyways */
     s_mqtt_client_connection_release_threadsafe_function(binding);
```
xiazhvera, line 107: If we are releasing the functions here, would we still need release in `s_on_connect_call`?

source/mqtt_client_connection.c
```diff
@@ -251,11 +272,115 @@ static void s_on_connection_resumed(
...
+        return;
+    }
+
+    struct connection_failure_args *args =
+        aws_mem_calloc(binding->allocator, 1, sizeof(struct connection_failure_args));
+    AWS_FATAL_ASSERT(args);
+    args->error_code = error_code;
+
+    AWS_NAPI_ENSURE(NULL, aws_napi_queue_threadsafe_function(binding->on_connection_failure, args));
+}
```
xiazhvera, line 377: We probably also wanna release the function after callback is called.

xiazhvera (changes requested):

source/mqtt_client_connection.c
alfred2g: this is asynchronous, so it might be deleted before the callback goes through no?

source/mqtt_client_connection.c
```diff
@@ -584,6 +747,9 @@ static void s_on_connect_call(napi_env env, napi_value on_connect, void *context
     if (args->return_code || args->error_code) {
         /* Failed to create a connection, none of the callbacks will be invoked again */
         s_mqtt_client_connection_release_threadsafe_function(binding);
+        if (binding->first_successfull_connection == true) {
```
xiazhvera, line 750: Looking into the code, it seems the `first_successfull_connection` will never be true here.Is it necessary then?

Other than that, PR looks good to me.

xiazhvera (approved):

source/mqtt_client_connection.c
```diff
@@ -45,8 +45,20 @@ struct mqtt_connection_binding {
     napi_threadsafe_function on_any_publish;
     napi_threadsafe_function transform_websocket;
     napi_threadsafe_function on_closed;
+    napi_threadsafe_function on_connection_success;
+    napi_threadsafe_function on_connection_failure;
+    bool first_successfull_connection;
```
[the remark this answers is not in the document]
alfred2g: it handles a first connection attempt, so if we are already connected before, the reconnected will take over, and no need to free the handles, else free the handles

sfod (approved):

PR #497, opened 2023-10-25
Fix signing leaks

graebm:

### DIFF
--- a/source/auth.c
+++ b/source/auth.c
@@ -18,7 +18,7 @@
 
 #include <aws/common/condition_variable.h>
 #include <aws/common/mutex.h>
-
+#include <aws/http/request_response.h>
 #include <aws/io/tls_channel_handler.h>
 
 static const char *AWS_NAPI_KEY_ENDPOINT = "endpoint";
@@ -584,7 +584,6 @@ done:
  **********************************************************************************************************************/
 
 struct signer_sign_request_state {
-    napi_ref node_request;
     struct aws_http_message *request;
     struct aws_signable *signable;
 
@@ -629,7 +628,7 @@ static void s_destroy_signing_binding(
     }
 
     /* Release references */
-    napi_delete_reference(env, binding->node_request);
+    aws_http_message_release(binding->request);
 
     const size_t num_blacklisted = binding->header_blacklist.length;
     for (size_t i = 0; i < num_blacklisted; ++i) {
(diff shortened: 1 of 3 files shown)


source/http_message.c
```diff
@@ -121,6 +121,10 @@ static void s_napi_http_request_finalize(napi_env env, void *finalize_data, void
     struct http_request_binding *binding = finalize_data;
     struct aws_allocator *allocator = finalize_hint;
 
+    if (binding->node_headers != NULL) {
+        napi_delete_reference(env, binding->node_headers);
```
graebm, line 125: Leak Fix 1: The C HttpRequest binding was never releasing its reference to the JS HttpHeaders

source/auth.c
```diff
@@ -641,7 +640,7 @@ static void s_destroy_signing_binding(
 
     aws_signable_destroy(binding->signable);
 
-    AWS_NAPI_ENSURE(env, aws_napi_unref_threadsafe_function(env, binding->on_complete));
+    AWS_NAPI_ENSURE(env, aws_napi_release_threadsafe_function(binding->on_complete, napi_tsfn_abort));
```
graebm, line 643: Leak Fix 2: release() decrements the reference count, not ~unref()~

source/auth.c
```diff
@@ -584,7 +584,6 @@ static napi_value s_creds_provider_new_x509(napi_env env, const struct aws_napi_
  **********************************************************************************************************************/
 
 struct signer_sign_request_state {
-    napi_ref node_request;
     struct aws_http_message *request;
```
graebm, line 587: Simplification/Theoretical Fix: Just keep refcount on the C request. Don't bother storing pointer to the node_request.

Previously, we kept a refcount on the node_request, and just had a pointer to the C request (without doing any refcounting). That kiiinda works for the aws_sign_request() path, but the verify_sigv4a_signing() path forgot to refcount the node_request and just did a pointer to the C request which is SKETCHY.

source/module.c
```diff
@@ -947,10 +947,10 @@ napi_status aws_napi_dispatch_threadsafe_function(
         call_status = status;
         s_handle_failed_callback(env, function, status);
     });
+    /* main thread can exit now */
+    napi_unref_threadsafe_function(env, tsfn);
     /* Must always decrement the ref count, or the function will be pinned */
     napi_status release_status = napi_release_threadsafe_function(tsfn, napi_tsfn_release);
-    /* main thread can exit now */
-    AWS_NAPI_ENSURE(env, napi_unref_threadsafe_function(env, tsfn));
     return (call_status != napi_ok) ? call_status : release_status;
```
graebm, line 954: Theoretical Fix: move unref() call to happen before the release(). [the docs say](https://nodejs.org/api/n-api.html#reference-counting-of-thread-safe-functions):
> napi_release_threadsafe_function() should be the last API call made in conjunction with a given napi_threadsafe_function, because after the call completes, there is no guarantee that the napi_threadsafe_function is still allocated

waahm7 (approved):

TingDaoK (approved):

PR #499, opened 2023-10-30
Node minimum version bump

bretambrose:
https://github.com/awslabs/aws-crt-nodejs/discussions/468

* Update minimum supported node version to 14
* Update cmake-js, crypto-js, and axios dependencies
* Don't disable eventstream callbacks until the user explicitly invokes close
* Re-enable racy eventstream test
* Misc fixups related to dependency updates


.github/workflows/ci.yml
```diff
@@ -30,7 +30,6 @@ jobs:
           - al2-x64
           - manylinux2014-aarch64
           - manylinux2014-x64
-          - manylinux2014-x86
```
sbSteveK, line 33: ooc, why are we removing this?

bretambrose: linux x86 builds of node don't exist from 14 on

PR #501, opened 2023-11-07
Fix Segfault on Electron Exit

xiazhvera:
*Issue #, if available:*

*Description of changes:*
As Electron would shutdown the node process on exit, and eventually causing "segfault" on threadsafe function operation. We now disable the node threadsafe functions on node process exit to prevent the segfault. 


lib/native/binding.js
```diff
@@ -74,5 +74,15 @@ if (binding == undefined) {
     throw new Error("AWS CRT binary not present in any of the following locations:\n\t" + search_paths.join('\n\t'));
 }
 
+import crt_native from "./binding"
+/** As Electron would shutdown the node process on exit, and eventually causing "segfault" on threadsafe function operation.
```
sbSteveK, line 78: Trivial: "Electron will shutdown the node process on exit, which causes the threadsafe function to segfault. To prevent the segfault we disable the threadsafe function on node process exit."

xiazhvera:
Though it is possible to put the rw lock on module data, I didn't find a good way to retrieve the module data when we do threadsafe function operations. One possible solution is to make the whole module data a global static, while I feel it was meaningless to do so...
I will setup a separate task for running test suite in Electron.

PR #503, opened 2023-11-15
Setup Short Run Mqtt5 Canary

xiazhvera:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/canary/mqtt5/canary.ts
+++ b/canary/mqtt5/canary.ts
@@ -25,17 +25,17 @@ yargs.command('*', false, (yargs: any) => {
             default: 'localhost',
         },
         'port': {
-            description: 'number: port to connect to',
+            description: 'INT: port to connect to',
             type: 'number',
             default: 1883,
         },
         'tps': {
-            description: 'number: transaction per second',
+            description: 'INT: transaction per second',
             type: 'number',
             default: 0,
         },
         'clients': {
-            description: 'number: concurrent running clients',
+            description: 'INT: concurrent running clients',
             type: 'number',
             default: 10,
         }
@@ -72,17 +72,16 @@ interface CanaryContext {
 
     mqttStats: CanaryMqttStatistics;
 
-    subscriptions: string[];
+    subscriptions: string[][];
 }
 
 function sleep(millisecond: number) {
     return new Promise((resolve) => setInterval(resolve, millisecond));
 }
 
-function getRandomClient(clients : mqtt5.Mqtt5Client[]): mqtt5.Mqtt5Client
+function getRandomIndex(clients : mqtt5.Mqtt5Client[]): number
 {
-    let index = Math.floor(Math.random() * clients.length);
-    return clients[index];
+    return Math.floor(Math.random() * clients.length);
 }
 
 function createCanaryClients(testContext: TestContext, mqttStats: CanaryMqttStatistics): mqtt5.Mqtt5Client[] {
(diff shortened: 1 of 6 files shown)


canary/mqtt5/canary.ts
```diff
@@ -90,8 +139,8 @@ async function doUnsubscribe(context : CanaryContext) {
     try {
         context.mqttStats.unsubscribesAttempted++;
 
-        await context.client.unsubscribe({
-            topicFilters: [ topicFilter ]
+        await getRandomClient(context.clients).unsubscribe({
```
sfod, line 142: We have a check above in this function that subscription exists. With multiple clients this logic seems not working anymore

canary/mqtt5/canary.ts
```diff
@@ -13,62 +13,111 @@ type Args = { [index: string]: any };
...
+            type: 'number',
+            default: 120,
+        },
+        'endpoint': {
+            description: 'STR: endpoint to connect to',
+            type: 'string',
+            default: 'localhost',
+        },
+        'port': {
+            description: 'number: port to connect to',
```
sfod, line 28: nit: duration option uses `INT` word in the description, while other numeric options use `number`. It's better to use one term, for consistency.

canary/mqtt5/canary.ts
```diff
@@ -101,119 +150,126 @@ async function doUnsubscribe(context : CanaryContext) {
...
+async function doPublish(context: CanaryContext, qos: mqtt5.QoS) {
     try {
         context.mqttStats.publishesAttempted++;
 
-        await context.client.publish({
+        // Generate random binary payload data
+        let payload = Buffer.alloc(10000);
+        for(let i = 0; i < 10000; i++)
+        {
+            payload[i] = Math.floor(Math.random() * 128);
```
sfod, line 161: A question: does initializing buffer with random bytes affect publishing in any way?

xiazhvera: I thought about it, and it is probably not necessary for canary. Using a random bytes for payload is more for "functional test". It can make sure we handle the string encoding correctly. However, it is not a required for canary. 

The payload size still matters, as a large payload size would make it easier to observe memory issue related to payload, but the payload data is not necessary a focus. 

I will revert it.

sfod: I see, thanks for the explanation!

canary/mqtt5/canary.ts
```diff
@@ -13,207 +13,263 @@ type Args = { [index: string]: any };
...
         });
 
-        context.subscriptions.push(topicFilter);
+        context.subscriptions[index].push(topicFilter);
         context.mqttStats.subscribesSucceeded++;
     } catch (err) {
         context.mqttStats.subscribesFailed++;
-        context.subscriptions.filter(entry => entry !== topicFilter);
+        if(context.subscriptions[index].length > 0 )
+            context.subscriptions[index].filter(entry => entry !== topicFilter);
```
sfod, line 129: It seems `filter` method doesn't change array, instead it creates a new array with entries that passed the filter:

```
let subscriptions = ["a", "b", "c"];
let filtered_subscriptions = subscriptions.filter(entry => entry !== "b"); 
console.log("subscriptions: " + subscriptions);
console.log("filtered_subscriptions: " + filtered_subscriptions);
```

```
subscriptions: a,b,c
filtered_subscriptions: a,c
```

But `context.subscriptions[index].push(topicFilter);` is at the very end, so filtering a new topic out is probably not needed.

canary/mqtt5/canary.ts
```diff
@@ -13,207 +13,263 @@ type Args = { [index: string]: any };
...
-            topicFilters: [ topicFilter ]
+        await context.clients[index].unsubscribe({
+            topicFilters: [topicFilter]
         });
 
         context.mqttStats.unsubscribesSucceeded++;
     } catch (err) {
         context.mqttStats.unsubscribesFailed++;
-        context.subscriptions.push(topicFilter);
+        context.subscriptions[index].push(topicFilter);
```
sfod, line 151: I think the check on the line 134 should be `if (context.subscriptions[index].length == 0)`. Otherwise, we're going to add to `context.subscriptions[index]` `"canthappen"` topicfilter.

sfod (approved):

PR #508, opened 2023-11-28
Hardcoded `napi_no_external_buffers_allowed` value for Node 21 and below

xiazhvera:
This reverts commit 0aa0feb63859b8f1f6bca1759249df1d74c15ef2.

*Issue #, if available:*

*Description of changes:*
The enum `napi_no_external_buffers_allowed` is introduced in node21 and backport to node 14.21.2, 16.19.0, 18.13.0.
(Details in : https://github.com/nodejs/node/pull/45181)
We hardcoded `napi_no_external_buffers_allowed` value until we upgrade to node21. 

### DIFF
--- a/source/module.c
+++ b/source/module.c
@@ -50,8 +50,9 @@ AWS_STATIC_ASSERT(NAPI_VERSION >= 4);
 
 /* TODO:
  * Hardcoded enum value for `napi_no_external_buffers_allowed`.
- * The enum `napi_no_external_buffers_allowed` is introduced in node14.
- * Use it for external buffer related changes after bump to node 14 */
+ * The enum `napi_no_external_buffers_allowed` is introduced in node21 and backport
+ * to node 14.21.2, 16.19.0, 18.13.0.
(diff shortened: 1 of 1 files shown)


source/module.c
```diff
@@ -50,8 +50,9 @@ AWS_STATIC_ASSERT(NAPI_VERSION >= 4);
 
 /* TODO:
  * Hardcoded enum value for `napi_no_external_buffers_allowed`.
- * The enum `napi_no_external_buffers_allowed` is introduced in node14.
- * Use it for external buffer related changes after bump to node 14 */
+ * The enum `napi_no_external_buffers_allowed` is introduced in node21 and backport
```
graebm, line 53: But I see it mentioned here in the node14 docs page?
https://nodejs.org/docs/latest-v14.x/api/n-api.html#n_api_napi_status

xiazhvera: The enum is back-ported to node 14.21.2, 16.19.0, 18.13.0. And it would not available for all version for node14,16,18.
While we could not know which version the customer would use, it would be safer to have the const instead of directly using the enum.

graebm (approved):

PR #509, opened 2023-11-29
Add shared subscriptions test

sfod:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/test/mqtt5.ts
+++ b/test/mqtt5.ts
@@ -576,3 +576,116 @@ export async function doRetainTest(client1: mqtt5.Mqtt5Client, client2: mqtt5.Mq
     await stopped1;
     client1.close();
 }
+
+export async function doSharedSubscriptionsTest(publisher: mqtt5.Mqtt5Client, subscriber1: mqtt5.Mqtt5Client, subscriber2: mqtt5.Mqtt5Client) {
+    const payload : Buffer = Buffer.from("share", "utf-8");
+    const messageNumber: number = 10;
+    const testTopic: string = `mqtt5_test${uuid()}`;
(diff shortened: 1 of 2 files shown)


test/mqtt5.ts
```diff
@@ -576,3 +576,116 @@ export async function doRetainTest(client1: mqtt5.Mqtt5Client, client2: mqtt5.Mq
...
+    let receivedResolve : (value?: void | PromiseLike<void>) => void;
+    let receivedPromise = new Promise<void>((resolve, reject) => {
+        receivedResolve = resolve;
+        setTimeout(() => reject(new Error("Did not receive expected number of messages")), 4000);
+    });
+
+    const clientsReceived = new Set();
+    const messagesReceived = new Map();
+
+    subscriber1.on('messageReceived', (eventData: mqtt5.MessageReceivedEvent) => {
```
bretambrose, line 625: Minor but a couple of things:

* The two closures are the same except for "sub1" vs "sub2". I bet there's a straightforward way of adding one more layer of indirection such that you could have a single large closure and two tiny ones that invoke it with the right parameter, removing the duplication.
* This feels like overkill since it spends effort verifying server behavior which is technically out of scope.  Feels like a single {clientId -> received_count} map and a Sum(received_count) + failure_count == sent_count assertion would better match the scope.

sfod: Thanks for the feedback, I addressed it except one thing:
`Sum(received_count) + failure_count == sent_count`
Is `failure_count` a number of failed publishes? If yes, I'm not sure if it's possible to get a failure and do not cause a timeout.

bretambrose (approved):

PR #513, opened 2023-12-05
Conditionally uri encode custom auth signature values

bretambrose:


sfod (approved):

bretambrose:
> The PR looks good to me. However, would it be better to have the encoding validation in aws-c-mqtt instead of on binding layer? as we had this issue from other language as well.

How would aws-c-mqtt know how to parse the username field?

PR #514, opened 2023-12-08
Don't let browser choose port in 311 implementation

bretambrose:


lib/browser/ws.ts
```diff
@@ -116,11 +116,11 @@ export function create_websocket_url(config: MqttConnectionConfig) {
         const day = canonical_day(time);
         const query_params = `X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=${signing_config.credentials.aws_access_id}` +
             `%2F${day}%2F${signing_config.credentials.aws_region}%2F${signing_config.service}%2Faws4_request&X-Amz-Date=${time}&X-Amz-SignedHeaders=host`;
-        const url = new URL(`wss://${config.host_name}${path}?${query_params}`);
+        const url = new URL(`wss://${config.host_name}:${config.port}${path}?${query_params}`);
```
xiazhvera, line 119: Would this break current user ?

bretambrose: Shouldn't.  The browser impl cannot connect to Iot Core on any port other than 443.

xiazhvera (approved):

PR #515, opened 2023-12-11
Misc fixes that should have been in the first PR

bretambrose:
* Remove encoding warnings in documentation
* encode authorizer name and token key name as well, per aws documentation

### DIFF
--- a/lib/common/aws_iot_shared.ts
+++ b/lib/common/aws_iot_shared.ts
@@ -112,8 +112,7 @@ export interface MqttConnectCustomAuthConfig {
     /**
      * Name of the custom authorizer to use.
      *
-     * Required if the endpoint does not have a default custom authorizer associated with it.  It is strongly suggested
-     * to URL-encode this value; the SDK will not do so for you.
+     * Required if the endpoint does not have a default custom authorizer associated with it.
      */
     authorizerName?: string;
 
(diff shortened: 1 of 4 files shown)


lib/common/aws_iot_shared.ts
```diff
@@ -157,36 +155,35 @@ export interface MqttConnectCustomAuthConfig {
...
+        return encodeURIComponent(value);
     }
 }
 
 /** @internal */
 export function canonicalizeCustomAuthConfig(config: MqttConnectCustomAuthConfig) : MqttConnectCustomAuthConfig {
     let processedConfig : MqttConnectCustomAuthConfig = {};
 
-    utils.set_defined_property(processedConfig, "authorizerName", config.authorizerName);
+    utils.set_defined_property(processedConfig, "authorizerName", canonicalizeCustomAuthEncoding(config.authorizerName));
```
xiazhvera, line 181: Question: regards to the authorizerName. As the canonicalizeCustomAuthEncoding  only checks for `%` , is it possible that username contains `%`? I thought we use `%`  to validate the signature because the signature is base64 encoded and should never include a `%`. Is that also true to authorizerName?

bretambrose: AWS documentation suggests uri-encoding all three of: token key name, token value signature, and authorizer name.  If an authorizer name did contain a % naturally (unsure if it's allowed) then it would have to be encoded as "%25" which is exactly what we do here.

xiazhvera: I think our `canonicalizeCustomAuthEncoding` would return the original value instead of encoding it if it contains a `%`. If `%` is allowed for authorizer name, it seems that we would not encode it here with `canonicalizeCustomAuthEncoding`.

bretambrose: Hmm you're right.  Need to rethink this.

PR #516, opened 2023-12-12
Documentation updates for custom auth connections

bretambrose:

### DIFF
--- a/lib/common/aws_iot_shared.ts
+++ b/lib/common/aws_iot_shared.ts
@@ -81,14 +81,19 @@ export function populate_username_string_with_custom_authorizer(
     if (is_string_and_not_empty(input_authorizer) && input_authorizer) {
         username_string = add_to_username_parameter(username_string, input_authorizer, "x-amz-customauthorizer-name=");
     }
+
+    if (is_string_and_not_empty(input_signature) || is_string_and_not_empty(input_token_value) || is_string_and_not_empty(input_token_key_name)) {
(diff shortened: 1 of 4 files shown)


lib/common/aws_iot_shared.ts
```diff
@@ -83,17 +83,11 @@ export function populate_username_string_with_custom_authorizer(
     }
     if (is_string_and_not_empty(input_signature) && input_signature) {
         username_string = add_to_username_parameter(username_string, input_signature, "x-amz-customauthorizer-signature=");
-        if ((is_string_and_not_empty(input_token_key_name) && input_token_key_name) || (is_string_and_not_empty(input_token_value) && input_token_value))
-        {
-            console.log("Warning: Signed custom authorizers with signature will not work without a token key name and " +
-                        "token value. Your connection may be rejected/stalled on the IoT Core side due to this. Please " +
-                        "set the token key name and token value to connect to a signed custom authorizer.");
-        }
```
sfod, line 91: That's a useful warning, no? It just need to be fixed: instead of `if (A or B)` should be `if (!A or !B)`

sfod: Okay, now I see that these parameters are validated below. Though, the situation when `input_token_value` and `input_token_key_name` are provided while `input_signature` is empty is not handled. Maybe add this check as well?

bretambrose: Sure

sfod (approved):

PR #490, opened 2023-10-03
Napi search path

bretambrose:
Add environment-injected fallback options for NAPI binary search path set

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #522, opened 2024-01-10
Fix "Cannot find module" for aws-crt-nodejs

ferbs:
*Description of changes:*
In `lib/native/bindings.js`, after successfully finding the aws-crt-nodejs native module, it drops the *.node* extension when calling require. That should be fine in a pure commonJs environment but dropping the extension breaks some ES-like runtime and build environments.

This PR includes the extension in the require call.

I don't see a test to update. (Consider adding some? There seem to be a lot of regressions in aws-sdk-js-v3 relating to aws-crt and signature-v4-crt. It would take a lot of work around test fixtures but perhaps worthwhile/justified?)

### DIFF
--- a/lib/native/binding.js
+++ b/lib/native/binding.js
@@ -63,12 +63,14 @@ const search_paths = [
 ];
 
 let binding;
-for (const path of search_paths) {
-    if (existsSync(path + '.node')) {
-        binding = require(path);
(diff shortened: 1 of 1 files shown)


bretambrose:
Should be superceded by https://github.com/awslabs/aws-crt-nodejs/pull/490

PR #525, opened 2024-01-18
Add win_delay_load_hook

xiazhvera:
*Issue #, if available:*

*Description of changes:*
This change will fix dll load issue on Windows. It would allow us to use Electron and vercel/pkg. 

### DIFF
--- a/source/windows/win_delay_load_hook.c
+++ b/source/windows/win_delay_load_hook.c
@@ -4,22 +4,21 @@
  */
 
 /*
- * This file sets up a windows delay load hook that the dll would redirect to 
+ * This file sets up a windows delay load hook that the dll would redirect to
  * the calling process (.exe file) instead of `node.exe`.
- * 
+ *
  * This allows compiled addons to work when node.exe is renamed.
  */
 
 #ifdef _MSC_VER
 
-#ifndef WIN32_LEAN_AND_MEAN
-#define WIN32_LEAN_AND_MEAN
-#endif
+#    ifndef WIN32_LEAN_AND_MEAN
+#        define WIN32_LEAN_AND_MEAN
+#    endif
 
-#include <windows.h>
-#include <delayimp.h>
-#include <string.h>
-
+#    include <delayimp.h>
+#    include <string.h>
+#    include <windows.h>
 
 FARPROC WINAPI delayHook(unsigned dliNotify, PDelayLoadInfo pdli) {
     switch (dliNotify) {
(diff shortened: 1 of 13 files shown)


bretambrose (commented):
I'm very uncomfortable with making a change that affects everyone and switches from C to C++.  I think we should push the maintainer of cmake-js to add a build option that would cause their copy of win_delay_load_hook to be treated as C.

The primary blocker there looks to be finding a replacement for the decltype shorthand in the last line.

xiazhvera:
> I'm very uncomfortable with making a change that affects everyone and switches from C to C++. I think we should push the maintainer of cmake-js to add a build option that would cause their copy of win_delay_load_hook to be treated as C.
> 
> The primary blocker there looks to be finding a replacement for the decltype shorthand in the last line. It sure seems like it could just be replaced by the actual function signature above since it appears to be a fixed ABI contract.

We can probably add our own win_delay_load_hook implementation instead of using cmakejs version. Let me update it.

bretambrose (commented):
Is there any way to add a CI step that makes and runs a simple electron exe on windows?

xiazhvera:
> Is there any way to add a CI step that makes and runs a simple electron exe on windows?

As all our samples are in js-v2 branch, the electron sample would also launched there:https://github.com/aws/aws-iot-device-sdk-js-v2/pull/398 (WIP, still need setup CI for it.)

I also created an quick electron sample to verify the change here:
https://github.com/awslabs/aws-crt-nodejs/pull/526. 
Checkout [windows](https://github.com/awslabs/aws-crt-nodejs/actions/runs/7590555985/job/20677261356#logs) and  [windows-vc14-x86](https://github.com/awslabs/aws-crt-nodejs/actions/runs/7590555985/job/20677261949?pr=526#logs)

bretambrose (approved):

source/windows/win_delay_load_hook.c
```diff
@@ -0,0 +1,66 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
```
bretambrose, line 5: Actually hold up.  We should make sure everything is kosher license-wise with this.

source/windows/win_delay_load_hook.c
```diff
@@ -0,0 +1,66 @@
...
+#    include <string.h>
+
+FARPROC WINAPI delayHook(unsigned dliNotify, PDelayLoadInfo pdli) {
+    switch (dliNotify) {
+        case dliNotePreLoadLibrary:
+
+            // If you want to return control to the helper, return 0.
+            // Otherwise, return your own HMODULE to be used by the
+            // helper instead of having it call LoadLibrary itself.
+            if (_stricmp(pdli->szDll, "node.exe") != 0) {
```
bretambrose, line 36: Also let's not use _stricmp

bretambrose (changes requested):

bretambrose (approved):

PR #540, opened 2024-03-19
Add polyfill for non-window browser contexts

MYoung25:
*Description of changes:*
This PR updates the browser polyfill check to check for any browser-based global context through `self` instead of window.  It should enable `aws-crt-nodejs` to work in web and service worker contexts.

------


MYoung25:
Meant to be a draft PR to share a concept

PR #544, opened 2024-04-23
Service Work Sample

xiazhvera:
*Issue #, if available:*

*Description of changes:*


bretambrose:
Any way we can get CI for this case?

xiazhvera:
> Any way we can get CI for this case?

We can potentially setup a service worker sample and have it run in CI. However, I'm hesitated about having a service worker sample live in our main repo, as we haven't officially declared support for service workers. Including the sample might lead to misunderstandings among users.
Instead, I would prefer making the sample available as a reference rather than an official example. This way, we can provide guidance. 
Let me know what you think.

PR #559, opened 2024-07-18
explicitly depend on ws 8.17.1 for nested dependencies to use

sbSteveK:
Explicitly set ws dependency to v8.17.1 for nested dependencies to use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "yargs": "^17.7.2"
   },
   "dependencies": {
+    "ws": "^8.17.1", // @httptoolkit/websocket-stream nested ws * dependencies need to use this version or higher.
     "@aws-sdk/util-utf8-browser": "^3.259.0",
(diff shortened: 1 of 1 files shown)


bretambrose:
That's not how nested dependencies work in JS unfortunately.

sbSteveK:
> That's not how nested dependencies work in JS unfortunately.

I could be misunderstanding something but here's where I'm at.

httptoolkit/websocket-stream and its dependency isomorphic-ws both have their ws dependencies set to version '*'
The nuxt dependency has its ws dependency set to ^8.17.0
This appears to be causing all ws depencencies in those 3 to be set to 8.17.0. I believe manually setting a ws dependency of 8.17.1 should deduplicate to 8.17.1 for those 3 dependencies.

mqtt is a separate issue as its ws dependency is a different major version (7.5.9). That being the case, this doesn't fix the ws dependency issue for mqtt.

PR #560, opened 2024-07-22
Simple logging API; defaults to console

bretambrose:
Needed for MQTT request-response browser implementation.  If reasonable, we should add lightweight logging to MQTT wrappers as well.

### DIFF
--- a/lib/common/io.ts
+++ b/lib/common/io.ts
@@ -63,3 +63,86 @@ export enum SocketDomain {
     /** UNIX domain socket/Windows named pipes */
     LOCAL = 2,
 }
+
+/**
+ * The amount of detail that will be logged
+ * @category Logging
+ */
+export enum LogLevel {
+    /** No logging whatsoever. */
+    NONE = 0,
+    /** Only fatals. In practice, this will not do much, as the process will log and then crash (intentionally) if a fatal condition occurs */
+    FATAL = 1,
+    /** Only errors */
+    ERROR = 2,
+    /** Only warnings and errors */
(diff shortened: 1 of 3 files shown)


lib/common/io.ts
```diff
@@ -63,3 +63,95 @@ export enum SocketDomain {
...
+ * @category Logging
+ */
+export enum LogLevel {
+    /** No logging whatsoever. */
+    NONE = 0,
+    /** Only fatals. In practice, this will not do much, as the process will log and then crash (intentionally) if a fatal condition occurs */
+    FATAL = 1,
+    /** Only errors */
+    ERROR = 2,
+    /** Only warnings and errors */
```
sbSteveK, line 78: Extremely Trivial: This is the only comment that explicitly calls out itself along with another log category. It's unlikely but this may cause someone to assume that later levels are not inclusive of previous log levels.

lib/common/io.ts
```diff
@@ -63,3 +63,95 @@ export enum SocketDomain {
...
+    DEBUG = 5,
+    /** Everything. Only use this if you really need to know EVERY single call */
+    TRACE = 6
+}
+
+let logLevel : LogLevel = LogLevel.NONE;
+
+/**
+ * Sets the amount of detail that will be logged
+ * @param level - maximum level of logging detail.  Log invocations at a higher level of detail will be ignored.
```
sbSteveK, line 92: Related: This deffo clears up the issue above for the most part.

sbSteveK (approved):

PR #561, opened 2024-07-29
Request response browser2

bretambrose:
* Subscription manager implementation and tests.  Both impl and tests are direct ports from the corresponding pieces of aws-c-mqtt


lib/browser/mqtt_request_response/protocol_adapter.ts
```diff
@@ -461,4 +469,17 @@ export class ProtocolClientAdapter extends BufferedEventEmitter {
                 return false;
         }
     }
+
+    private static isSubackReasonCodeRetryable(reasonCode: mqtt5.SubackReasonCode) : boolean {
+        switch (reasonCode) {
+            case mqtt5.SubackReasonCode.UnspecifiedError:
```
sbSteveK, line 475: Debatable: UnspecifiedError seems like something we don't want to retry. Seems like the kind of thing where a retry wouldn't resolve but I don't have strong enough feelings about it if you put it here for a reason.

lib/browser/mqtt_request_response/subscription_manager.ts
```diff
@@ -0,0 +1,635 @@
...
+        this.adapter.addListener(protocol_adapter.ProtocolClientAdapter.SUBSCRIBE_COMPLETION, this.handleSubscribeCompletionEvent.bind(this));
+        this.adapter.addListener(protocol_adapter.ProtocolClientAdapter.UNSUBSCRIBE_COMPLETION, this.handleUnsubscribeCompletionEvent.bind(this));
+        this.adapter.addListener(protocol_adapter.ProtocolClientAdapter.CONNECTION_STATUS, this.handleConnectionStatusEvent.bind(this));
+    }
+
+    close() {
+        this.closed = true;
+
+        this.unsubscribeAll();
+    }
```
sfod, line 191: Not sure if this is an issue or not, but I think, there is a possibility for a subscription to succeed after the `close` is called. Then, that subscription will remain active after the manager is closed.

Here is a hypothetical case:
1. User calls `manager.acquireSubscription("a/b")` --> `this.activateSubscription("a/b")` --> `this.adapter.subscribe("a/b")`
2. User calls `manager.close()` --> `this.unsubscribe("a/b")` --> `this.adapter.unsubscribe("a/b")`
3. MQTT broker processes `SUBSCRIBE` and returns a retryable SUBACK --> `this.activateSubscription("a/b")` --> `this.adapter.subscribe("a/b")`
4. MQTT broker processes `UNSUBSCRIBE`.
5. MQTT broker processes `SUBSCRIBE` with a success.

Even without the `retryable` logic, it's possible that MQTT broker will process subscribe and unsubscribe requests in reversed order:
3. MQTT broker processes `UNSUBSCRIBE`
4. MQTT broker processes `SUBSCRIBE`
But I don't think we can handle this in a robust way.

But the retryable logic probably could be improved by checking `closed` flag.

bretambrose: Predicating the resubscribe with a check of this.closed would address, yes?

Beyond that I am unsure how the broker would receive an unsubscribe and a subscribe out of submission order.

sfod: > Predicating the resubscribe with a check of this.closed would address, yes?

I believe yes, it should help.

lib/browser/mqtt_request_response/protocol_adapter.ts
bretambrose: Removed

lib/browser/mqtt_request_response/protocol_adapter.ts
```diff
@@ -461,4 +469,17 @@ export class ProtocolClientAdapter extends BufferedEventEmitter {
                 return false;
```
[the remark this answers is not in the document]
bretambrose: Seems like a good idea.

PR #562, opened 2024-08-05
Mqtt js upgrade

bretambrose:
Waiting on 
* New MQTTJS release to fix unsubscribe typing
* s2n config solution so that aws-lc is used during feature probes


lib/browser/mqtt.ts
```diff
@@ -310,7 +311,7 @@ export class MqttClientConnection extends BufferedEventEmitter {
 
         const will = this.config.will ? {
             topic: this.config.will.topic,
-            payload: normalize_payload(this.config.will.payload),
+            payload: normalize_payload_to_buffer(this.config.will.payload),
```
xiazhvera, line 314: This seems to be a breaking change where `normalize_payload_to_buffer` always return a buffer, while `normalize_payload` could return any type. Should we use `normalize_payload_to_buffer` for all payload in that case?

lib/browser/mqtt.ts
```diff
@@ -576,7 +577,8 @@ export class MqttClientConnection extends BufferedEventEmitter {
                     return this.on_error(error);
                 }
                 const sub = (packet as mqtt.ISubscriptionGrant[])[0];
-                resolve({ topic: sub.topic, qos: sub.qos });
+                // MV TOFIX: 128 can be passed here and is not modeled in QoS
+                resolve({ topic: sub.topic, qos: sub.qos as QoS });
```
xiazhvera, line 581: Ideally, it should not be an issue as we would throw error before we reach here when returned qos is 128.

lib/browser/mqtt.ts
xiazhvera: Never mind, my fault. The will would only be accessed internally, it should not affect any user.

xiazhvera (approved):

PR #566, opened 2024-08-09
Use node-16 for Mac OS arm64

sfod:
*Issue #, if available:*

macos job currently fails in CI.

*Description of changes:*

Prior to Node v16, Mac OS arm64 has only experimental support (for example, see [Node v15 docs](https://github.com/nodejs/node/blob/v15.x/BUILDING.md#platform-list)). So, using node-16 for Mac OS arm64 job seems the right solution.

### DIFF
--- a/builder.json
+++ b/builder.json
@@ -68,9 +68,9 @@
             "!imports": [ "node-14" ]
         },
         "macos": {
-            "_comment": "Mac OS (arm64) has only experimental support for Node v15 and earlier, so stick to Node v16, see Node v15 docs: https://github.com/nodejs/node/blob/v15.x/BUILDING.md#platform-list",
             "architectures": {
                 "arm64": {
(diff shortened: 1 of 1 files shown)


builder.json
```diff
@@ -66,6 +66,14 @@
         "al2012": {
             "!packages": [],
             "!imports": [ "node-14" ]
+        },
+        "macos": {
+            "_comment": "Mac OS (arm64) has only experimental support for Node v15 and earlier, so stick to Node v16, see Node v15 docs: https://github.com/nodejs/node/blob/v15.x/BUILDING.md#platform-list",
```
TingDaoK, line 71: trivial: move the comment within the `arm64` architectures, since it's specific to arm64 Mac.

TingDaoK (approved):

PR #573, opened 2024-08-16
Continuous Delivery nodejs version bump

sbSteveK:
Bump the install phase nodejs version from 10 to 14 in continuous-delivery related yml files.

CD is currently failing at the POST_BUILD phase of the aws-crt-nodejs-cd-combine-binaries CodeBuild Build Project with the context "COMMAND_EXECUTION_ERROR: Error while executing command: bash ./continuous-delivery/pack.sh. Reason: exit status 1" 

logs indicate:
npx tsc -p tsconfig.json
Unexpected token ?

run(`npx tsc -p tsconfig.json`) is in the `scripts/tsc.js`
Which is listed in the "prepare": "node ./scripts/tsc.js && node ./scripts/install.js" of the package.json.

The related yml file for pack sets the runtime-version of nodejs to 10. Changing this to our minimum supported version of 14 should resolve the issue. We may as well also bump the version from 10 -> 14 in the publish and test-version-exists files as well since I'm expecting the same issue in those locations.

This solution was tested locally where running `npx tsc -p tsconfig.json` resulted in no error when using node 14 but generated an identical "Unexpected token ?" error when using Node 10.

### DIFF
--- a/continuous-delivery/pack.yml
+++ b/continuous-delivery/pack.yml
@@ -2,7 +2,7 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      nodejs: 10
+      nodejs: 14
   pre_build:
     commands:
       - cd aws-crt-nodejs
--- a/continuous-delivery/publish.yml
+++ b/continuous-delivery/publish.yml
@@ -2,7 +2,7 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      nodejs: 10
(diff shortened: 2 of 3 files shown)


graebm:
Why?

Please, in all pull requests explain WHY you're doing things. Explain the problem you encountered, and why this will fix it.

graebm (approved):

PR #583, opened 2024-10-01
Bind out crc64 and fix various hash bidning issues

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*


lib/browser/crypto.ts
```diff
@@ -12,25 +12,34 @@
...
+ * CryptoJS does not provide easy access to underlying bytes.
+ * As a workaround just dump it to a string and then reinterpret chars as individual bytes.
+ * TODO: long term we would probably want to move to WebCrypto for SHA's and some other 3p for crc's and md5.
+ * @param hash 
+ * @returns 
  */
-export class Md5Hash {
-    private hash?: Crypto.WordArray;
+function hashToUint8Array(hash: Crypto.WordArray) {
+    return Uint8Array.from(hash.toString(Crypto.enc.Latin1).split('').map(c => c.charCodeAt(0)));;
```
TingDaoK, line 27: why do we use `Crypto.enc.Latin1`?

https://github.com/brix/crypto-js/issues/274
Looks like there were some discussions about this.

package.json
```diff
@@ -23,15 +23,15 @@
     "test:browser:ci": "npm run install:puppeteer && npm run test:browser",
     "install:puppeteer": "npm install --save-dev jest-puppeteer puppeteer @types/puppeteer",
     "prepare": "node ./scripts/tsc.js && node ./scripts/install.js",
-    "install": "node ./scripts/install.js"
+    "install": "node ./scripts/install.js --rebuild"
```
TingDaoK, line 26: is this necessary?

DmitriyMusatkin: no, will remove. i was experimenting with something and forgot to remove.

lib/browser/crypto.ts
DmitriyMusatkin: Its mostly for simplicity purposes. Dumping to hex and then parsing from it is slightly more code. 
I tried out parsing the array out of the words initally and it was not that much of perf difference and would require updating types dependency, which breaks mqtt code. Instead of mucking further with cryptoJS and its deps at that point i think it would be better to spend that time on dropping cryptoJS altogether, so i just went with whatever works with current deps.

source/checksums.c
```diff
@@ -55,9 +53,69 @@ napi_value crc_common(napi_env env, napi_callback_info info, uint32_t (*checksum
...
+    uint8_t *buffer = to_hash.buffer;
+    size_t length = to_hash.len;
+
+    uint64_t previous = 0;
+    if (!aws_napi_is_null_or_undefined(env, node_args[1])) {
+        if (aws_byte_buf_init_from_napi(&previous_buf, env, node_args[1])) {
+            napi_throw_type_error(env, NULL, "previous argument must be undefined or a positive number");
+            goto done;
+        }
+        AWS_FATAL_ASSERT(previous_buf.len == 8);
```
TingDaoK, line 95: should we check the length and throw the type error instead of crashing?

The node function takes a DataView, which could be anything. It's probably something we want to properly handle.

source/checksums.c
```diff
@@ -55,9 +53,69 @@ napi_value crc_common(napi_env env, napi_callback_info info, uint32_t (*checksum
...
+        napi_throw_type_error(env, NULL, "to_hash argument must be a string or array");
+        goto done;
+    }
+    uint8_t *buffer = to_hash.buffer;
+    size_t length = to_hash.len;
+
+    uint64_t previous = 0;
+    if (!aws_napi_is_null_or_undefined(env, node_args[1])) {
+        if (aws_byte_buf_init_from_napi(&previous_buf, env, node_args[1])) {
+            napi_throw_type_error(env, NULL, "previous argument must be undefined or a positive number");
```
TingDaoK, line 92: The `previous` is confusing to be a name of the variable or a real word. 

```suggestion
            napi_throw_type_error(env, NULL, "`previous` argument must be undefined or a positive number");
```

Also, a positive number doesn't seem to descriptive enough as a 64bit data? Oh, I guess you were copying from the crc32 case. So, we should update this.

lib/browser/crypto.ts
TingDaoK: Maybe renaming to `wordArrayToUint8Array`?

Maybe add some comments about how `Crypto.enc.Latin1` and `charCodeAt` to convert the buffer back and forth? 

So, firstly, I thought we should use base64 as we commonly used. But, after digging, it seems not a directly match when converted to string. And the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt) of `charCodeAt` said it used `utf-16`?

Agreed on whatever works, but I feel those encoding are really confusing.

source/checksums.c
DmitriyMusatkin: yep. added this for test purposes. let me replace it with proper error.

source/checksums.c
DmitriyMusatkin: yeah, previous naming was copied from crc32 case. let me update the naming to indicate its previous crc

lib/browser/crypto.ts
DmitriyMusatkin: completely agree and i think eventually it makes sense to move off CryptoJS and use WebCrypto as well. 
current complication is that webcrypto only has shas and we need alternatives for crc's and md5 (if we want to keep a browser version).
I added a comment, but the jist of it is that latin1 is 8bit encoding (my understanding its the same as 8bit ascii variant) and charCodeAt just retrieves the corresponding byte value (it can return values bigger than a byte for unicode, but in this case we are limited by chars not being unicode to byte size values)

source/checksums.c
```diff
@@ -30,22 +34,16 @@ napi_value crc_common(napi_env env, napi_callback_info info, uint32_t (*checksum
...
     uint8_t *buffer = to_hash.buffer;
     size_t length = to_hash.len;
-    uint32_t previous = 0;
+    uint32_t previous_crc = 0;
 
     if (!aws_napi_is_null_or_undefined(env, node_args[1])) {
-        if (napi_get_value_uint32(env, node_args[1], &previous)) {
-            napi_throw_type_error(env, NULL, "previous argument must be undefined or a positive number");
+        if (napi_get_value_uint32(env, node_args[1], &previous_crc)) {
+            napi_throw_type_error(env, NULL, "previous value argument must be undefined or a positive number");
```
TingDaoK, line 41: I think this error is meant to point the variable name in nodejs function [here](https://github.com/awslabs/aws-crt-nodejs/pull/583/files#diff-21a8783999cf20978be916d17129b69a5a45f60d7d20102cf25b7681501ad17bR51)

Maybe rename the nodejs variable name to `previous_crc`, and use `previous_crc` here as well?

source/checksums.c
```diff
@@ -55,9 +53,72 @@ napi_value crc_common(napi_env env, napi_callback_info info, uint32_t (*checksum
...
+    uint64_t previous_crc = 0;
+    if (!aws_napi_is_null_or_undefined(env, node_args[1])) {
+        if (aws_byte_buf_init_from_napi(&previous_crc_buf, env, node_args[1])) {
+            napi_throw_type_error(env, NULL, "previous value argument must be undefined or a DataView over 8 bytes");
+            goto done;
+        }
+        if (previous_crc_buf.len != 8) {
+            napi_throw_type_error(env, NULL, "previous value argument must be a DataView over 8 bytes");
+            goto done;
+        }
```
TingDaoK, line 98: trivial:
```suggestion
        if (aws_byte_buf_init_from_napi(&previous_crc_buf, env, node_args[1]) || previous_crc_buf.len != 8) {
            napi_throw_type_error(env, NULL, "previous value argument must be undefined or a DataView over 8 bytes");
            goto done;
        }
```

TingDaoK (commented):
just trivial comments

source/checksums.c
DmitriyMusatkin: im on a fence about it. previous_crc is a local variable that does not have much meaning to caller in js land. from that perspective i think just referring to it as previous value makes more sense

source/checksums.c
DmitriyMusatkin: i actually initially wrote it that way, but merging those 2 ifs looked uglier to me, so i broke them apart for clarity, but happy to discuss

source/checksums.c
TingDaoK: mostly because all the other place we have the error throw out is referring to the variable name used in the nodejs land. I think it's better to be consistent

TingDaoK (approved):

PR #584, opened 2024-10-10
Prebuild S2N

TingDaoK:
*Issue #, if available:*
- Prebuild S2N and update submodule for using s2n_cleanup_thread()

*Description of changes:*

### DIFF
--- a/.builder/actions/crt_size_check.py
+++ b/.builder/actions/crt_size_check.py
@@ -74,5 +74,4 @@ class CrtSizeCheck(Builder.Action):
 
         print(f"Total NPM package file size: {total_size} bytes")
         if total_size > max_size:
-            raise Exception(f'NPM package exceeds size limit of {
-                            max_size} bytes.')
+            raise Exception(f'NPM package exceeds size limit of {max_size} bytes.')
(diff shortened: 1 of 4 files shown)


.builder/actions/crt_size_check.py
```diff
@@ -11,7 +11,7 @@ def run(self, env):
         # Maximum package size (for current platform) in bytes
         # NOTE: if you increase this, you might also need to increase the
         # limit in continuous-delivery/pack.sh
-        max_size = 7_500_000
+        max_size = 8_000_000
```
waahm7, line 14: Do we need to increase the limit in pack.sh?

TingDaoK: don't know... the pack.sh is not exactly match the CI here, you could try to run the test pipeline see if we need to

waahm7: We will find out

waahm7 (approved):

.builder/actions/crt_size_check.py
waahm7: We found out in #587

PR #590, opened 2024-11-25
Logger fallback to console.error if process._rawDebug is missing

nickg:
When running under Deno instead of NodeJS I get the following stack trace when the native extension is loaded:

```
N-API call failed: napi_create_threadsafe_function( env, node_rawdebug, NULL, resource_name, 0, 1, ctx, s_threadsafe_log_finalize, ctx, s_threadsafe_log_call, &ctx->log_drain)                                                                                                                                                                           
    @ /codebuild/output/src1849732470/src/aws-crt-nodejs/source/logger.c:277: napi_function_expected                                                   
Fatal error condition occurred in /codebuild/output/src1849732470/src/aws-crt-nodejs/source/logger.c:277: napi_create_threadsafe_function( env, node_rawdebug, NULL, resource
_name, 0, 1, ctx, s_threadsafe_log_finalize, ctx, s_threadsafe_log_call, &ctx->log_drain)                                                                                
Exiting Application
```

Deno does not provide `console._rawDebug` so fall back to `console.error` instead.

### DIFF
--- a/source/logger.c
+++ b/source/logger.c
@@ -258,6 +258,21 @@ void s_threadsafe_log_create(struct aws_napi_logger_ctx *ctx, napi_env env) {
     napi_value node_rawdebug = NULL;
     AWS_NAPI_ENSURE(env, napi_get_named_property(env, node_process, "_rawDebug", &node_rawdebug));
 
+    napi_valuetype rawdebug_type;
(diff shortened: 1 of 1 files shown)


waahm7:
Thanks, following up in a new PR to fix the lint.

PR #592, opened 2024-12-16
Logger fallback to console.error if process._rawDebug is missing

waahm7:
Duplicate of https://github.com/awslabs/aws-crt-nodejs/pull/590.
 When running under Deno instead of NodeJS I get the following stack trace when the native extension is loaded:

```
N-API call failed: napi_create_threadsafe_function( env, node_rawdebug, NULL, resource_name, 0, 1, ctx, s_threadsafe_log_finalize, ctx, s_threadsafe_log_call, &ctx->log_drain)                                                                                                                                                                           
    @ /codebuild/output/src1849732470/src/aws-crt-nodejs/source/logger.c:277: napi_function_expected                                                   
Fatal error condition occurred in /codebuild/output/src1849732470/src/aws-crt-nodejs/source/logger.c:277: napi_create_threadsafe_function( env, node_rawdebug, NULL, resource
_name, 0, 1, ctx, s_threadsafe_log_finalize, ctx, s_threadsafe_log_call, &ctx->log_drain)                                                                                
Exiting Application
```

Deno does not provide `console._rawDebug` so fall back to `console.error` instead.

### DIFF
--- a/source/logger.c
+++ b/source/logger.c
@@ -264,13 +264,13 @@ void s_threadsafe_log_create(struct aws_napi_logger_ctx *ctx, napi_env env) {
     /* process._rawDebug is specific to NodeJS and may not exist in
        other environments like Deno, fall back to console.error */
     if (rawdebug_type == napi_undefined) {
-       napi_value node_console = NULL;
-       AWS_NAPI_ENSURE(env, napi_get_named_property(env, node_global, "console", &node_console));
+        napi_value node_console = NULL;
+        AWS_NAPI_ENSURE(env, napi_get_named_property(env, node_global, "console", &node_console));
 
-       napi_value node_error = NULL;
-       AWS_NAPI_ENSURE(env, napi_get_named_property(env, node_console, "error", &node_error));
+        napi_value node_error = NULL;
+        AWS_NAPI_ENSURE(env, napi_get_named_property(env, node_console, "error", &node_error));
 
-       node_rawdebug = node_error;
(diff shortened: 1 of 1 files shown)


bretambrose:
I don't think we should merge PRs containing code that never gets meaningfully exercised anywhere in testing or CI.  I'm also not thrilled about adding support for non-Node execution engines.

waahm7:
@bretambrose 
> I don't think we should merge PRs containing code that never gets meaningfully exercised anywhere in testing or CI. I'm also not thrilled about adding support for non-Node execution engines.

In general, I agree, but for this particular change, `_rawDebug` is a private function that we should not be using from the Node code [[link](https://github.com/nodejs/node/blob/b171afefb61c6c0f54fe1421438ffffbdd0c9a33/src/node_process.h#L24)].  
```
// Most of the time, it's best to use `console.error` to write  
// to the `process.stderr` stream. However, in some cases, such as  
// when debugging the `stream.Writable` class or the `process.nextTick`  
// function, it is useful to bypass JavaScript entirely.
```
I don't know the Node/JavaScript world well, but I think `console.error` is recommended over `process._rawDebug`. Since `process._rawDebug` is an internal function, Node is free to remove or change its implementation in the future, so it seems reasonable to fall back to `console.error` if it is not found. I haven't made `console.error` the default due to performance concerns and not knowing enough to make this change.

I have tested the code in CI to ensure that everything works even if we always use `console.error`. However, I haven't tested it locally using Deno.

PR #596, opened 2025-01-21
Fix memory leak during signing

graebm:
**Issue:** (internal ticket V1637484435)

When doing SigV4A signing from aws-sdk-js-v3, memory usage gradually increases due to memory leaks.

**Description of changes:**
Fix leak that occurred if headers were queried repeatedly.

**Future Work:**
All native unit tests should check that they aren't leaking memory. I spent a few hours trying to add such tests, but couldn't get it working in a reliable way. I didn't want this to hold back the fix, so leaving it for another day. See these blog posts for hints on various techniques NodeJS has used for leak testing through the years: https://joyeecheung.github.io/blog/2024/03/17/memory-leak-testing-v8-node-js-1/, https://joyeecheung.github.io/blog/2024/03/17/memory-leak-testing-v8-node-js-2/

### DIFF
--- a/source/http_connection.c
+++ b/source/http_connection.c
@@ -149,8 +149,11 @@ struct http_connection_binding {
 /* finalizer called when node cleans up this object */
 static void s_http_connection_from_manager_binding_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
     (void)finalize_hint;
-    (void)env;
     struct http_connection_binding *binding = finalize_data;
+
+    if (binding->node_external != NULL) {
+        napi_delete_reference(env, binding->node_external);
+    }
 
     /* no release call, the http_client_connection_manager has already released it */
     aws_mem_release(binding->allocator, binding);
--- a/source/http_connection_manager.c
+++ b/source/http_connection_manager.c
@@ -27,8 +27,10 @@ struct aws_http_connection_manager *aws_napi_get_http_connection_manager(
 
 static void s_http_connection_manager_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
     (void)finalize_hint;
-    (void)env;
     struct http_connection_manager_binding *binding = finalize_data;
+    if (binding->node_external != NULL) {
+        napi_delete_reference(env, binding->node_external);
+    }
     aws_mem_release(binding->allocator, binding);
 }
 
--- a/source/http_message.c
+++ b/source/http_message.c
@@ -79,19 +79,20 @@ napi_status aws_napi_http_message_bind(napi_env env, napi_value exports) {
 
 struct http_request_binding {
     struct aws_http_message *native;
-    struct aws_allocator *allocator;
 
     napi_ref node_headers;
 };
 
-/* Need a special finalizer to avoid releasing a request object we don't own */
-static void s_napi_wrapped_http_request_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
-    (void)env;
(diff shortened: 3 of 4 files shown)


source/http_connection.c
```diff
@@ -149,9 +149,12 @@ struct http_connection_binding {
 /* finalizer called when node cleans up this object */
 static void s_http_connection_from_manager_binding_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
     (void)finalize_hint;
-    (void)env;
     struct http_connection_binding *binding = finalize_data;
 
+    if (binding->node_external != NULL) {
+        napi_delete_reference(env, binding->node_external);
+    }
```
graebm, line 156: This fixes another leak I found when searching for uses of `napi_create_reference()/napi_delete_reference()`)

No one is using the NodeJS bindings of our HTTP classes, as far as I know...

source/http_connection_manager.c
```diff
@@ -27,8 +27,10 @@ struct aws_http_connection_manager *aws_napi_get_http_connection_manager(
 
 static void s_http_connection_manager_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
     (void)finalize_hint;
-    (void)env;
     struct http_connection_manager_binding *binding = finalize_data;
+    if (binding->node_external != NULL) {
+        napi_delete_reference(env, binding->node_external);
+    }
```
graebm, line 33: This fixes another leak I found when searching for uses of `napi_create_reference()/napi_delete_reference()`)

No one is using the NodeJS bindings of our HTTP classes, as far as I know...

source/http_message.c
```diff
@@ -115,20 +115,6 @@ struct aws_http_message *aws_napi_http_message_unwrap(napi_env env, napi_value j
  * Constructor
  **********************************************************************************************************************/
 
-static void s_napi_http_request_finalize(napi_env env, void *finalize_data, void *finalize_hint) {
```
graebm, line 118: There were 2 different finalizers for the same class. I combined them into 1.

(When this code was first written, the 2 different finalizers were necessary because the underlying C classes didn't have reference-counting. But now they do, so the weirdness the led to needing 2 different finalizers isn't necessary anymore)

source/http_message.c
```diff
@@ -238,15 +224,16 @@ static napi_value s_request_headers_get(napi_env env, void *native_this) {
...
+         * Create a Node object, and a reference to it, and store that reference on this HTTP request */
         struct aws_http_headers *headers = aws_http_message_get_headers(binding->native);
         AWS_NAPI_ENSURE(env, aws_napi_http_headers_wrap(env, headers, &result));
+        AWS_NAPI_ENSURE(env, napi_create_reference(env, result, 1, &binding->node_headers));
     }
 
-    /* Store the value for later */
-    AWS_NAPI_ENSURE(env, napi_create_reference(env, result, 1, &binding->node_headers));
-
     return result;
```
graebm, line 237: This was the leak. We were creating a new `napi_ref` each time the headers were queried. Now, we only create a `napi_ref` if we didn't already have one.

source/http_stream.c
```diff
@@ -74,7 +74,7 @@ static void s_on_response_call(napi_env env, napi_value on_response, void *conte
     }
 
     /* clean up the response buffer */
-    aws_http_message_destroy(binding->response);
+    aws_http_message_release(binding->response);
```
graebm, line 77: this isn't a bugfix, just replacing a deprecated _destroy() function with a modern _release() function

waahm7 (approved):

PR #611, opened 2025-04-11
Add required keepAliveIntervalSeconds in MQTT5 PubSub React sample

MakendranG:
*Issue #, if available:*
N/A

*Description of changes:*
Added required keepAliveIntervalSeconds parameter in the MQTT5 PubSub React sample as it's a mandatory field according to: https://github.com/awslabs/aws-crt-nodejs/blame/main/lib/common/mqtt5_packet.ts#L878

Changes made:
- Added keepAliveIntervalSeconds: 30 to the connect packet configuration in PubSub React sample
- This parameter is required as per MQTT5 specification and cannot be optional

This fixes the TypeScript compilation error:
TS2345: Argument of type '{ clientId: string; }' is not assignable to parameter of type 'ConnectPacket'.
Property 'keepAliveIntervalSeconds' is missing in type '{ clientId: string; }' but required in type 'ConnectPacket'.

### DIFF
--- a/lib/common/mqtt5_packet.ts
+++ b/lib/common/mqtt5_packet.ts
@@ -875,7 +875,7 @@ export interface ConnectPacket extends IPacket {
      *
      * See [MQTT5 Keep Alive](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901045)
      */
-    keepAliveIntervalSeconds: number;
+    keepAliveIntervalSeconds: 30;
 
     /**
      * A unique string identifying the client to the server.  Used to restore session state between connections.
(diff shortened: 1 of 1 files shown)


lib/common/mqtt5_packet.ts
```diff
@@ -875,7 +875,7 @@ export interface ConnectPacket extends IPacket {
      *
      * See [MQTT5 Keep Alive](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901045)
      */
-    keepAliveIntervalSeconds: number;
+    keepAliveIntervalSeconds: 30; // This is required, not optional
```
bretambrose, line 878: Please add this to the sample instead of the default.

MakendranG: Raised the same - https://github.com/aws/aws-iot-device-sdk-js-v2/pull/551

PR #613, opened 2025-04-25
Build with Warning as Error

sbSteveK:
windows-vc14-x86 was reporting a conversion from `size_t` to `uint32_t` with a possible loss of data when used with `napi_get_element`. 

In CI, we previously ran `npm install` before `node scripts/build.js -DAWS_WARNINGS_ARE_ERRORS=ON`. So that the first step actually build the shared lib without warning as errors and the next step just picked the cached built artifacts instead of building the source code with warning as errors, which hide the warnings generated by compiler.

Move the build with warnings as errors before npm install to make sure it reports warnings as error correctly.

### DIFF
--- a/builder.json
+++ b/builder.json
@@ -74,6 +74,20 @@
                     "!imports": [ "node-16" ]
                 }
             }
+        },
+        "windows": {
+            "!build_steps": [
+                [
+                    "node",
+                    "scripts/build.js",
+                    "-DAWS_WARNINGS_ARE_ERRORS=ON",
+                ],
+                [
+                    "npm",
+                    "install",
+                    "--unsafe-perm"
+                ]
+            ]
         }
     },
     "targets": {
--- a/builder.json
+++ b/builder.json
@@ -104,11 +104,6 @@
             "--unsafe-perm"
         ],
         [
-            "node",
-            "scripts/build.js",
-            "-DAWS_WARNINGS_ARE_ERRORS=ON"
-        ],
-        [
             "npm",
             "run-script",
             "tsc"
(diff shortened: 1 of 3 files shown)


TingDaoK (commented):
> This warning was previously being treated as an error but is no longer doing so

why? Can we keep the warning as error?

builder.json
```diff
@@ -74,6 +74,20 @@
...
+                    "node",
+                    "scripts/build.js",
+                    "-DAWS_WARNINGS_ARE_ERRORS=ON"
+                ],
+                [
+                    "npm",
+                    "install",
+                    "--unsafe-perm"
+                ]
+            ]
```
TingDaoK, line 90: Can we update `build_steps` at line 109 to run the `node scripts/build.js -DAWS_WARNINGS_ARE_ERRORS=ON` before `npm install`?
In the original PR, Mike said it related to the arm mac, but I guess i don't know exactly why. Maybe it's not the problem now??

Try it out, and see if we have any issues. Since it makes more sense to have all platform to take the warnings as errors.

sbSteveK: Seems to be good. Let me know if there's anything else.

builder.json
```diff
@@ -93,6 +93,11 @@
         ]
     ],
     "build_steps": [
+        [
+            "node",
+            "scripts/build.js",
+            "-DAWS_WARNINGS_ARE_ERRORS=ON"
+        ],
```
TingDaoK, line 100: remove the same step couple lines below, we don't need it now, right?

TingDaoK (approved):

TingDaoK:
also, rename the PR and commit message to something like `build with warning as error`, it doesn't really relate to flaky CI,

PR #645, opened 2025-10-01
Mqtt tls policy

bretambrose:


lib/native/binding.d.ts
```diff
@@ -55,6 +55,10 @@ export function is_alpn_available(): boolean;
 export function io_client_bootstrap_new(): NativeHandle;
 /* wraps aws_tls_context #TODO: Wrap with ClassBinder */
 /** @internal */
+export function io_tls_cipher_preference_is_supported(
```
sbSteveK, line 58: Should we not bind this in aws-crt-nodejs?

The only point where `tls_cipher_preference_is_supported` is used is in tests and the tests run the same check when creating a new tls ctx as it's the first thing investigated in `s_tls_ctx_new()` in aws-c-io. The tests should continue to pass and fail as expected with it removed.

PR #680, opened 2026-01-07
Add IoT Metrics Support

xiazhvera:
*Issue #, if available:*

*Description of changes:*
- Bind `IoTDeviceSDKMetrics` and metrics related features , CRT is now appending AWS metrics by default.
- Creating a shared base interface for MqttConnectionConfig and Mqtt5ClientOption

- Node: Add `disableMetrics` option to allow user disable metrics.
   - MQTT3: Add node-only `disable_metrics` option in MqttConnectionConfig 
   - MQTT5: Add node-only `disableMetrics` option in Mqtt5ClientOption

- Browser: 
  - renamed SDK metrics:  "SDK=NodeJSv2" => "SDK=IoTDeviceSDK/JS"


lib/common/aws_iot_shared.ts
```diff
@@ -201,10 +201,11 @@ function addParam(paramName: string, paramValue: string | undefined, paramSet: [
  * and SDK metrics properties.
  *
  * @param customAuthConfig intended AWS IoT custom auth client configuration
+ * @param appendMetrics manually append SDK metrics, set to true while we build username for browser.
```
bretambrose: This doesn't seem accurate.  There is no additional parameter.

lib/common/mqtt_shared.ts
```diff
@@ -62,6 +62,17 @@ export function normalize_payload_to_buffer(payload: any): Buffer {
...
 
+/** 
+ * IoT Device SDK Metrics Structure 
+ * @internal
+*/
+export class AwsIoTDeviceSDKMetrics {
+    /**
+     * Name of the library
+     */
+    libraryName: string = "IoTDeviceSDK/JS";
```
bretambrose: Can this just be a string constant somewhere and then all the tests and updates ref the string constant?

lib/native/mqtt.spec.ts
```diff
@@ -82,12 +82,40 @@ test_env.conditional_test(test_env.AWS_IOT_ENV.mqtt311_is_valid_direct_auth_mqtt
...
             password: test_env.AWS_IOT_ENV.MQTT311_BASIC_AUTH_PASSWORD,
-            socket_options: new SocketOptions()
+            socket_options: new SocketOptions(),
+            enable_metrics: false
         }
         await test_connection(config, new MqttClient(new ClientBootstrap()));
     })
 });
 
+test_env.conditional_test(test_env.AWS_IOT_ENV.mqtt311_is_valid_direct_auth_mqtt())('MQTT311 Connection - basic auth with metrics', async () => {
```
bretambrose, line 92: Can we do this against IoT Core; as it stands this will only get run in codebuild

lib/native/mqtt5.spec.ts
```diff
@@ -45,12 +45,15 @@ function createNodeSpecificTestConfig (testType: test_utils.SuccessfulConnection
             HttpProxyConnectionType.Tunneling);
     }
 
+    let enableMetrics = test_utils.ClientEnvironmentalConfig.isTestBasicAuth(testType) == false
```
bretambrose: metrics on if not basic auth but then the test with it enabled is basic auth?  what am I missing?

source/mqtt5_client.c
```diff
@@ -2328,6 +2335,19 @@ napi_value aws_napi_mqtt5_client_new(napi_env env, napi_callback_info info) {
         client_options.http_proxy_options = aws_napi_get_http_proxy_options(proxy_binding);
     }
 
+    // Set metrics
+    napi_value node_metrics = *arg++;
+    if (!aws_napi_is_null_or_undefined(env, node_metrics)) {
+        napi_value node_libraryName = NULL;
+        if (napi_get_named_property(env, node_metrics, "libraryName", &node_libraryName) == napi_ok &&
```
bretambrose: make "libraryName" a string constant at top of file

source/mqtt_client_connection.c
```diff
@@ -687,6 +690,31 @@ napi_value aws_napi_mqtt_client_connection_new(napi_env env, napi_callback_info
...
+    bool set_metrics_success = false;
+    if (!aws_napi_is_null_or_undefined(env, node_metrics)) {
+        napi_value node_libraryName = NULL;
+        if (napi_get_named_property(env, node_metrics, "libraryName", &node_libraryName) == napi_ok &&
+            aws_byte_buf_init_from_napi(&libraryName, env, node_libraryName) == AWS_OP_SUCCESS) {
+            metrics.library_name = aws_byte_cursor_from_buf(&libraryName);
+            set_metrics_success = true;
+        } else {
+            AWS_LOGF_DEBUG(AWS_LS_NODEJS_CRT_GENERAL, "Failed to retrieve SDK metrics");
+        }
```
bretambrose, line 709: shouldn't this go to cleanup?

source/mqtt_client_connection.c
```diff
@@ -687,6 +690,31 @@ napi_value aws_napi_mqtt_client_connection_new(napi_env env, napi_callback_info
...
+    }
+
+    if (aws_mqtt_client_connection_set_metrics(binding->connection, set_metrics_success ? &metrics : NULL)) {
+        AWS_LOGF_DEBUG(
+            AWS_LS_NODEJS_CRT_GENERAL,
+            "Failed to set metrics with error code %d(%s)",
+            aws_last_error(),
+            aws_error_debug_str(aws_last_error()));
+    }
```
bretambrose, line 718: go to cleanup?

source/mqtt_client_connection.c
```diff
@@ -687,6 +690,31 @@ napi_value aws_napi_mqtt_client_connection_new(napi_env env, napi_callback_info
...
 
+    // Set metrics
+    napi_value node_metrics = *arg++;
+    struct aws_mqtt_iot_metrics metrics;
+    AWS_ZERO_STRUCT(metrics);
+
+    bool set_metrics_success = false;
+    if (!aws_napi_is_null_or_undefined(env, node_metrics)) {
+        napi_value node_libraryName = NULL;
+        if (napi_get_named_property(env, node_metrics, "libraryName", &node_libraryName) == napi_ok &&
```
bretambrose: Named constant

lib/native/mqtt.spec.ts
xiazhvera: The test cannot run against IoT Core, as IoT Core will rule out the metrics string. The test actually uses the connection failure to verify that the metrics are appended.

xiazhvera:
TODO: as we have a mock server in place with the nodejs change, we will modify the test to use the mock server so we can actually capture the package and verify that the metrics are appended propelry.

source/mqtt_client_connection.c
xiazhvera: We don't fail the client if set metrics failed, we simply ignored the metrics in this case.

lib/native/mqtt.ts
```diff
@@ -217,6 +217,12 @@ export interface MqttConnectionConfig {
     /** Optional proxy options */
     proxy_options?: HttpProxyOptions;
 
+    /** Optional enable Aws IoT SDK Metrics. The metrics includes SDK name, version, and platform.
+     * 
+     * @group Node-only
```
bretambrose: Remove annotation given it will be shared.  Considering creating a base interface here with the shared values.

lib/native/mqtt5.ts
```diff
@@ -257,6 +257,14 @@ export interface Mqtt5ClientConfig {
      * @group Node-only
      */
     extendedValidationAndFlowControlOptions? : ClientExtendedValidationAndFlowControl;
+
+
+    /**
+     * Options for enable/disable Aws IoT Metrics. The metrics includes SDK name, version, and platform.
+     * 
+     * @group Node-only
```
bretambrose, line 213: Same, might as well remove

source/mqtt_client_connection.c
bretambrose: Why?  If we fail to do something that we've been asked to do, we usually propagate the failure rather than silently log.  What are the possible failure cases?

xiazhvera: A possible failure would be the username+metrics size is too long. For such case, I think the client should not prevent user from doing a successful connection if the original username itself is valid and within the size limit.

bretambrose (approved):

PR #694, opened 2026-03-03
Js mqtt - PR #1 Preliminary Foundation

bretambrose:
Fixes and updates to support an internal pure JS MQTT client implementation:

* Heap implementation to support operation timeout tracking
* Refactor reconnect wait calculations to be shared since the new client will be doing the same thing
* Accurate type annotations for our websocket connectors
* Fix a bug in buffered event emitter that would cause incorrect behavior if cork/uncork was used multiple times
* Add MQTT 311 reason codes to the MQTT5 reason code enums to support a client that can handle both protocols
* Add/refactor shared topic and topic filter analysis functions to support what the client will need


lib/browser/heap.spec.ts
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+import * as heap from "./heap";
+
+interface ElementType {
+    timestamp: number;
+    id: number;
```
xiazhvera, line 10: The `id` does not seem to be used in the tests.

lib/common/mqtt5_packet.ts
```diff
@@ -30,6 +30,32 @@ export enum ConnectReasonCode {
      */
     Success = 0,
 
+    /**
+     * (MQTT 311 Only) The connection was rejected because the server does not support the requested protocol version.
+     */
+    UnacceptableProtocolVersion311 = 1,
```
xiazhvera, line 36: I'm not sure what's the plan to support 2 protocols in one client. I will comeback after I go through the follow up PRs.

lib/browser/heap.spec.ts
bretambrose: It's the sort key.

bretambrose: Updated to verify that it breaks ties.

.builder/actions/crt_size_check.py
```diff
@@ -11,7 +11,7 @@ def run(self, env):
         # Maximum package size (for current platform) in bytes
         # NOTE: if you increase this, you might also need to increase the
         # limit in continuous-delivery/pack.sh
```
sbSteveK, line 13: Note: pack.sh currently set to 35,000 KB. This may or may not be high enough to cover the 2,000,000 k bump we make here. Might be safer to also bump pack.sh. I guess we'll find out when we run CD.

sbSteveK (approved):

xiazhvera:

PR #695, opened 2026-03-03
JS Mqtt - PR #2 Data model and packet logging

bretambrose:
Additional packet definitions and helper functions for the internal JS MQTT client.  Beyond what we currently expose in mqtt5_packet, we add two additional models:
1. An internal model that extends the public model of mqtt5_packet with fields we don't want to publicly expose (packet id, clean start, etc...),  We decode into this model.
1. A binary model (bytes or numbers only) for all operations, internal and external.  We do this to ensure we only pay the binary conversion cost of strings to bytes once.  This is the model we encode and write to the socket.

Also includes logging related functionality:
1. Additional helper functions for function-based logging that require a log level check pass before generating the log string
1. Helper functions for logging object properties to support packet logging


lib/common/io.ts
```diff
@@ -102,56 +102,71 @@ export function setLogLevel(level: LogLevel) {
...
 }
 
 export function logTrace(subject: string, logLine: string) {
-    if (logLevel < LogLevel.TRACE) {
+    log(LogLevel.TRACE, subject, logLine);
+}
+
+export type LogLineGenerator = () => string;
+
+function flog(level: LogLevel, subject: string, logLineGenerator: LogLineGenerator) {
```
xiazhvera, line 141: Considering that the `LogLineGenerator` returns a string, why would we need `flog` along side with `log` function?

lib/browser/mqtt_internal/model.ts
```diff
@@ -0,0 +1,1034 @@
...
+        case mqtt5_packet.PacketType.Pingreq:
+            return {
+                type: mqtt5_packet.PacketType.Pingreq
+            };
+        default:
+            throw new CrtError(`Unexpected/unsupported internal packet type: ${packet.type}`);
+    }
+}
+
+export function cloneSubscribeShallow(subscribe: mqtt5_packet.SubscribePacket) : mqtt5_packet.SubscribePacket {
```
xiazhvera, line 680: Why would we need the shallow clone here? Looking into the follow up PRs, we do a `convertPacketToBinary` immediately afterwards. It seems the clone is not necessary.

bretambrose: we use the original packet in a few places later (not in the same callstack) so I have it cloned so that the object can't change underneath.

lib/common/io.ts
bretambrose: flog skips string formatting if the log level check fails.  This lets us do expensive logging (like an entire packet and all its fields) and not pay the cost if the log level isn't met.

lib/browser/mqtt_internal/model.ts
```diff
@@ -0,0 +1,1034 @@
...
+export const CONNECT_5_PROTOCOL_BYTES = [0x00, 0x04, 0x4D, 0x51, 0x54, 0x54, 0x05];
+export const CONNECT_5_PROTOCOL_BUFFER = new Uint8Array(CONNECT_5_PROTOCOL_BYTES);
+export const CONNECT_5_PROTOCOL_DATAVIEW = new DataView(CONNECT_5_PROTOCOL_BUFFER.buffer);
+
+export const CONNECT_FLAGS_HAS_USERNAME : number = 0x80;
+export const CONNECT_FLAGS_HAS_PASSWORD : number = 0x40;
+export const CONNECT_FLAGS_HAS_WILL : number = 0x04;
+export const CONNECT_FLAGS_QOS_SHIFT : number = 0x03;
+export const CONNECT_FLAGS_WILL_RETAIN : number = 0x20;
+export const CONNECT_FLAGS_CLEAN_SESSION : number = 0x02;
```
sbSteveK, line 57: Just griping to the void (you) that MQTT5 spec names this flag `Clean Start` and MQTT3.1.1 names it `Clean Session`

lib/browser/mqtt_internal/model.ts
```diff
@@ -0,0 +1,1034 @@
...
+    return current;
+}
+
+function connectPacketToLogString(packet: ConnectPacketInternal, prefix : string) : string {
+    let result = `${prefix}Connect: {\n`;
+
+    result = log.appendBooleanPropertyLine(result, prefix, "CleanStart", packet.cleanStart);
+    result = log.appendNumericPropertyLine(result, prefix, "KeepAliveInterval", packet.keepAliveIntervalSeconds);
+    result = log.appendOptionalStringPropertyLine(result, prefix, "ClientId", packet.clientId);
+    // keep username opaque intentionally; there can be authentication-sensitive data in it
```
sbSteveK, line 815: trival: should we provide byte length of the username and password. If we are being opaque, should we not provide that and just log that it's present instead?

sbSteveK (approved):

lib/browser/mqtt_internal/model.ts
bretambrose: I don't think that adds any security.  This is mostly a helper for people that don't scrub log files.  Knowing the length of something doesn't give you a strong advantage in a brute-force attack.  It's way more expensive to check all N-length combinations than all 1..N-1 length combinations, so you don't gain much knowing the length.

xiazhvera:

PR #696, opened 2026-03-03
JS Mqtt - PR #3 Vli integer support

bretambrose:
Helper functions for encoding and decoding VLI integers based on the MQTT spec.


lib/browser/mqtt_internal/vli.ts
```diff
@@ -0,0 +1,81 @@
...
+    nextOffset?: number
+}
+
+export function decodeVli(data: DataView, offset: number) : VliDecodeResult {
+    let value: number = 0;
+    let index: number = 0;
+    let shift: number = 0;
+    while (index < 4) {
+        let view_index = offset + index++;
+        let raw_byte = data.getUint8(view_index);
```
sbSteveK, line 72: I think the view_index could potentially be outside the bounds of the DataView. We should do the `view_index+1>=data.byteLength` check before we run `data.getUint8()` on the index.

sbSteveK: This is looking locally here. Maybe there are checks in place where we are making sure the offset is valid when this function is called. Still, couldn't hurt to add the check just in case. Assuming things are going bad elsewhere if the offset is too high anyway.

sbSteveK (approved):
Potentially trivial check on out of bounds index due to offset. Besides that looks good.

lib/browser/mqtt_internal/vli.ts
bretambrose: A check could be added at the top of the function before the loop that there's at least one byte left to look at.

bretambrose: Added several offset checks and added some additional out-of-bounds or insufficient data tests

xiazhvera:

PR #699, opened 2026-03-03
JS Mqtt - PR #4 Packet encode and decode

bretambrose:
*Issue #, if available:*

*Description of changes:*


lib/browser/mqtt_internal/decoder.ts
```diff
@@ -0,0 +1,839 @@
...
+
+    throw new CrtError("insufficient data to decode variable-length integer");
+}
+
+export function decodeLengthPrefixedString(payload: DataView, offset: number) : [string, number] {
+    let [stringLength, index] = decodeU16(payload, offset);
+    return [toUtf8(new Uint8Array(payload.buffer, index, stringLength)), index + stringLength];
+}
+
+//Buffer.from(arrayBuffer, view.byteOffset, view.byteLength)
```
sbSteveK: Trivial: Can't tell if this and the two `// return` comments below were meant to convey some kind of information of if they need to be cleaned up.

lib/browser/mqtt_internal/decoder.ts
```diff
@@ -0,0 +1,839 @@
...
+    [value, index] = decodeLengthPrefixedString(payload, index);
+
+    userProperties.push({name: name, value: value});
+
+    return index;
+}
+
+// MQTT 311 Packet decoding functions
+
+function decodeConnackPacket311(firstByte: number, payload: DataView) : model.ConnackPacketInternal {
```
sbSteveK, line 75: trivial: We are verifying the `firstByte` in the connack for MQTT5 but not here. Feels pretty unnecessary but I don't see any reason not to verify the first byte for CONNACK is `PACKET_TYPE_FIRST_BYTE_CONNACK` here as well.

lib/browser/mqtt_internal/decoder.ts
```diff
@@ -0,0 +1,839 @@
...
+    let flags : number = 0;
+
+    let connack: model.ConnackPacketInternal = {
+        type: mqtt5_packet.PacketType.Connack,
+        sessionPresent: false,
+        reasonCode: mqtt5_packet.ConnectReasonCode.Success
+    };
+
+    [flags, index] = decodeU8(payload, index);
+    connack.sessionPresent = (flags & model.CONNACK_FLAGS_SESSION_PRESENT) != 0;
```
sbSteveK, line 317: Trivial: Nitpicking again but in the 311 connack we make sure that the reserved bits in the Acknowledge Flags are 0. Here we let it slide. We should stay consistent with ourselves in throwing for invalid flags.

lib/browser/mqtt_internal/decoder.ts
```diff
@@ -0,0 +1,839 @@
...
+        // @ts-ignore
+        return [result.value, result.nextOffset];
+    }
+
+    throw new CrtError("insufficient data to decode variable-length integer");
+}
+
+export function decodeLengthPrefixedString(payload: DataView, offset: number) : [string, number] {
+    let [stringLength, index] = decodeU16(payload, offset);
+    return [toUtf8(new Uint8Array(payload.buffer, index, stringLength)), index + stringLength];
```
sbSteveK, line 47: This kicked my brain and sent me down a rabbit hole of our MQTT5 encode/decode. I think we verify UTF-8 for encoding but not decoding despite spec stating various fields (topic, reason strings, etc) need to be UTF-8.

Maybe you remember something I don't about whether we do utf-8 validation on incoming packets. I'm not gonna spend more time digging through whether `toUtf8()` does validation if this doesn't matter but since we simply return here, unsure if we would report invalid utf-8 if the various payloads we feed it isn't actually utf-8.

sbSteveK: Just kidding. I did more digging.

The `@aws-sdk/util-utf8-browser` was deprecated and we are told to use https://npmx.dev/package/@aws-sdk/util-utf8 which was ALSO deprecated. 

Now we should use https://npmx.dev/package/@smithy/util-utf8 (spoke with Trivikram and there's not explicit reason but he suggests we not use the deprecated version).

This looks like it decodes and fails silently too so I dunno. Maybe there's no action to be taken here. We could probably parse the output string for U+FFFD (replacement character) since it's not allowed by the spec as well as what invalid characters would be replaced with but is it worth it? Probably not.

lib/browser/mqtt_internal/decoder.ts
bretambrose: Don't think that was intentional

lib/browser/mqtt_internal/decoder.ts
bretambrose: We can make moving to a different package a small task, but I don't think I want to do it here mid-stream

lib/browser/mqtt_internal/encoder.ts
```diff
@@ -0,0 +1,1066 @@
...
+                [mqtt5_packet.PacketType.Pingreq, (steps, packet) => { encodePingreqPacket(steps); }],
+            ]);
+
+    }
+
+    throw new CrtError("Unsupported protocol");
+}
+
+function addEncodingSteps(encoders: EncodingFunctionSet, steps: Array<EncodingStep>, packet: model.IPacketBinary) {
+    if (steps.length > 0) {
```
sbSteveK, line 922: trivial: this function is only called from one place and `this.steps` is always set to 0 there so this check seems uneccessary.

lib/browser/mqtt_internal/encoder.ts
```diff
@@ -0,0 +1,1066 @@
...
+        case mqtt5_packet.PacketType.Subscribe:
+            remaining_length = getSubscribePacketRemainingLengths311(packet as model.SubscribePacketBinary);
+            break;
+
+        case mqtt5_packet.PacketType.Unsubscribe:
+            remaining_length = getUnsubscribePacketRemainingLengths311(packet as model.UnsubscribePacketBinary);
+            break;
+
+        case mqtt5_packet.PacketType.Puback:
+            return 4; // packet type byte + remaining length (2) + packet id
```
sbSteveK, line 822: nitpick: below the number in the () represents the byte length but here it appears to be used as the byte position? Or it's an incorrect byte length. remaining length should be 1 byte and the packet id 2 bytes. Total is still correctly 4.

bretambrose: In both cases it's the value.

lib/browser/mqtt_internal/encoder.ts
bretambrose: Doesn't seem right to drop invariant checks just because the places you use something don't violate those invariants.

lib/browser/mqtt_internal/encoder.ts
```diff
@@ -0,0 +1,1066 @@
...
+    let will_properties_length: number = computeWillPropertiesLength(packet);
+
+    remaining_length += vli.getVliByteLength(properties_length) + properties_length;
+
+    if (packet.clientId) {
+        remaining_length += packet.clientId.byteLength;
+    }
+
+    if (packet.will) {
+        remaining_length += vli.getVliByteLength(will_properties_length) + will_properties_length
```
sbSteveK: Trivial: I'm assuming this is fine because it compiles but there's no `;` at the end of this statement.

bretambrose: It's fine but consistency is better.

lib/browser/mqtt_internal/encoder.ts
```diff
@@ -0,0 +1,1066 @@
...
+        case EncodingStepType.VLI:
+            return {
+                nextBuffer: vli.encodeVli(buffer, step.value as number)
+            };
+
+        case EncodingStepType.BYTES:
+            let source = step.value as DataView;
+            let amountToCopy = Math.min(buffer.byteLength, source.byteLength);
+
+            const destArray = new Uint8Array(buffer.buffer, buffer.byteOffset);
```
sbSteveK, line 972: Comment: Just below `Uint8Array()` is constructed with buffer, byteOffset, and length but here we don't provide a length.

Couldn't find any real info on whether creating a zero-length typed array is bad in any way but if we have the byteLength, maybe we should use it to be explicit? It's possible this thing defaults to the length of the buffer provided and just above we are doing a `Math.min()` between `buffer` and `source`. Could be we have a larger array created than needed. Trivial but just noting because of the difference in construction between the two `Uint8Array`s here.

sbSteveK (approved):
Some trivial comments that can be addressed or ignored. Looks solid

lib/browser/mqtt_internal/encoder.ts
bretambrose: From the docs

```
if only length is omitted, the remainder of buffer starting from byteOffset is viewed
```

It's important to note that these are not allocating memory, they're instancing specific views (cursors) over existing memory.

sbSteveK (approved):

xiazhvera:

PR #701, opened 2026-03-03
JS Mqtt - PR #5 Packet validation support

bretambrose:
Adds validation functions for user-submitted packets and broker-received packets.


lib/browser/mqtt_internal/validate.ts
```diff
@@ -0,0 +1,728 @@
...
+            case mqtt5_packet.SubackReasonCode.WildcardSubscriptionsNotSupported:
+                break;
+
+            default:
+                throw new CrtError(`"${value}" is not a valid MQTT5 SubackReasonCode`);
+        }
+    }
+}
+
+function validatePubackReasonCode(value: mqtt5_packet.PubackReasonCode, mode: model.ProtocolMode) {
```
sbSteveK: Trivial: Unsure why mode is here if it's not used.

lib/browser/mqtt_internal/validate.ts
```diff
@@ -0,0 +1,728 @@
...
+
+    for (let userProperty of userProperties) {
+        validateBufferLength(userProperty.name, "UserProperty.name");
+        validateBufferLength(userProperty.value, "UserProperty.value");
+    }
+}
+
+function validateBinaryPublish(packet: model.PublishPacketBinary, mode: model.ProtocolMode, settings: mqtt5_common.NegotiatedSettings, isWill: boolean) {
+    if (isWill) {
+        validatePacketLength(packet, mode, 65535);
```
sbSteveK: Trivial: Maybe it doesn't matter and nobody will pass along a will this large but assuming this limit came from the restriction on the size of a utf-8 string in the mqtt spec, I think both the topic and payload could each individually contain this size. If this is testing that them combined cannot exceed 65535 we may be failing validation on something that the spec accepts.

sbSteveK (approved):
Trivial nitpicks. Address as you see fit.

xiazhvera:

PR #702, opened 2026-03-03
JS Mqtt - PR #6 Protocol implementation

bretambrose:
Implementation of the client-side protocol requirements for MQTT 311 and 5


lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2200 @@
...
+    private buildConnectPacket() : model.ConnectPacketInternal {
+        let connectOptions = this.config.connectOptions;
+
+        let connect_packet : model.ConnectPacketInternal = {
+            type: mqtt5_packet.PacketType.Connect,
+            keepAliveIntervalSeconds: connectOptions.keepAliveIntervalSeconds,
+            cleanStart: this.computeCleanStart()
+        };
+
+        if (connectOptions.clientId != undefined) {
```
sbSteveK, line 1991: If a clientId isn't provided, we should generate one and set it here to avoid the reconnect issue where AWS provides a non-reusable clientId.

Could be generated elsewhere or further up but this seems like an okay place to generate one.

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2206 @@
...
+
+        this.pendingWriteCompletionOperations = [];
+
+        this.pendingWriteCompletion = false;
+    }
+
+    /**
+     * Helper function to determine if the clean start flag should be set on the client's Connect packet
+     */
+    private computeCleanStart() : boolean {
```
sbSteveK, line 1967: The enum for ResumeSessionPolicyType states that Default is the same as PostSuccess but here default is returning true while PostSuccess is checking against hasSuccessfullyConnected. It feels like a case for Never should return true and default should follow the case for PostSuccess.

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2206 @@
...
+                return;
+            }
+
+            try {
+                validate.validateInboundPacket(packet, this.config.protocolVersion);
+            } catch (e) {
+                flogError(PROTOCOL_STATE_LOG_SUBJECT, () => `Invalid packet of type ${mqtt5_packet.PacketType[packet.type ?? -1]} received`);
+
+                this.halt(HaltEventType.ProtocolError, e as CrtError);
+                continue;
```
sbSteveK: Just checking whether we want to continue processing packets after we called `this.halt()` because when handling incoming data we immediately eject if `this.haltState` is defined. I think we may want to cease processing packets but we might be continuing on post a failed validate for a reason?

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2206 @@
...
+                this.pendingPingrespTimeoutElapsedMillis = this.elapsedMillis + timeoutMillis;
+
+                flogInfo(PROTOCOL_STATE_LOG_SUBJECT, () => `Pingresp timeout in ${timeoutMillis} millis`);
+            }
+        }
+
+        // check for and handle a pingresp timeout
+        if (this.pendingPingrespTimeoutElapsedMillis) {
+            if (this.elapsedMillis >= this.pendingPingrespTimeoutElapsedMillis) {
+                logInfo(PROTOCOL_STATE_LOG_SUBJECT, "Pingresp timeout exceede.");
```
sbSteveK: Trivial: "exceeded"

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2206 @@
...
+
+        try {
+            // perform initial user-packet validation
+            validate.validateInitialOutboundPacket(packet, this.config.protocolVersion);
+
+            flogInfo(PROTOCOL_STATE_LOG_SUBJECT, () => `Submitting operation ${operation.id} of type ${mqtt5_packet.PacketType[operation.type]}`);
+            // don't want to log the packet until it passes initial validation
+            flogDebug(PROTOCOL_STATE_LOG_SUBJECT, () => model.internalPacketToLogString(packet, "  "));
+
+            // ofline queue policy check
```
sbSteveK: Utterly Trivial: "offline"

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2206 @@
...
+    /** Value to use for the maximum packet size property in the Connect packet */
+    maximumPacketSizeBytes?: number;
+
+    /** Value to use for the will delay interval property in the Connect packet */
+    willDelayIntervalSeconds?: number;
+
+    /** Value to use for the will property in the Connect packet */
+    will?: mqtt5_packet.PublishPacket;
+
+    /** u\User properties to use in the Connect packet */
```
sbSteveK: " u\" ?

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2206 @@
...
+                return (this.highPriorityOperationQueue.length > 0) ? this.elapsedMillis : undefined;
+
+            case ServiceQueueType.All:
+                if (this.highPriorityOperationQueue.length > 0 || this.resubmitOperationQueue.length > 0) {
+                    return this.elapsedMillis;
+                }
+
+                // user operations require more logic since there are certain conditions that we cannot
+                // dequeue them (receive maximum limit, no packet ids, etc...)
+                let id = this.userOperationQueue.shift();
```
sbSteveK, line 1096: Trivial: Instead of `shift()` `unshift()` maybe we should use `let id = this.userOperationQueue[0]`
https://stackoverflow.com/questions/42501871/peek-operation-in-stack-using-javascript
This should return undefined if nothing is there and we can check against `this.canDequeueUserOperation` without the need to unshift it back in place.

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2206 @@
...
+
+    private encoder : encoder.Encoder;
+    private decoder : decoder.Decoder;
+
+    private pendingWriteCompletion : boolean = false;
+    private pendingWriteCompletionOperations : Array<number> = new Array<number>();
+    private pendingFlushOperations: Array<number> = new Array<number>();
+
+    private pendingPublishAcks : Map<number, number> = new Map<number, number>();
+    private pendingNonPublishAcks : Map<number, number> = new Map<number, number>();
```
sbSteveK: Why do non publish have a separate pending queue? Looking around it appears as though on a reconnect the pendingPublishAcks get prioritized over pendingNonPublishAcks. If we have a series like
1 - subscribe goes out
2 -  publish to subscribed topic goes out
3 - disconnect -> `handleConnectionClosed() is run`
   a - `pendingNonPublishAcks` populates `userOperationQueue`
   b - `pendingPublishAcks` populates `resubmitOperationQueue`
4 - resume session
5 - in `dequeueNextOperation()` called with serviceQueueType ALL we go through high, then resubmit, and then finally userOperations.

This can result in an out of order where subscribe or unsubscribe goes out AFTER a publish that is using the topic of the sub/unsub resulting in either getting or not getting a publish they were or were not meant to get.

This might be trivial and I don't know off-hand if we follow this behavior in our native client but it feels wrong. We should probably order all QoS 1 resume session operations in order to happen on a rejoining of session.

sbSteveK (commented):
Mainly trivial but one area of concern in regard to rejoining a session operation order.

lib/browser/mqtt_internal/protocol.ts
bretambrose: default: and `ResumeSessionPolicyType.Default` are not the same thing

lib/browser/mqtt_internal/protocol.ts
bretambrose: Unsure, don't have a strong opinion here atm.

lib/browser/mqtt_internal/protocol.ts
bretambrose: I thought the spec explicitly forbids resending sub/unsubs ahead of interrupted qos1+ publishes but on re-reading it only requires the publishes to be in the same order, so we can merge these two.

lib/browser/mqtt_internal/protocol.ts
sbSteveK: Ah, cool. If `case ResumeSessionPolicy.PostSuccess:` also handles `ResumeSessionPolicyType.Default` because they both equate to `0` then no issue

lib/browser/mqtt_internal/protocol.ts
bretambrose: I'm not comfortable using an array as a filterable dequeue (example is a strict stack) using indexing operations.

lib/browser/mqtt_internal/protocol.ts
bretambrose: Looking closer, I agree with your suggestion.  Inbound validation is all major protocol requirements (packet ids, reason codes, etc...) as opposed to minor (packet size, etc...). So a failure here represents a significant error, on par with a decoding failure.  I'll early out.

sbSteveK (approved):

lib/browser/mqtt_internal/protocol.ts
```diff
@@ -0,0 +1,2228 @@
...
+    private serviceOutboundOperations(serviceQueueType : ServiceQueueType, socketBuffer: ArrayBuffer) : ServiceResult {
+        logDebug(PROTOCOL_STATE_LOG_SUBJECT, "Begin servicing outbound operations");
+
+        let done : boolean = false;
+        let remainingView = new DataView(socketBuffer);
+
+        while (!done) {
+            let currentOperation : ClientOperation | undefined = undefined;
+            if (this.currentOperation != undefined) {
+                currentOperation = this.operations.get(this.currentOperation);
```
xiazhvera, line 1347: It's a little confusing that `this.currentOperation` is the operations id ,while `currentOperation` is the operation object. Maybe rename it to `this.currentOperationId`?

xiazhvera:

PR #703, opened 2026-03-03
Js Mqtt - PR #7 Client implementation

bretambrose:
Client implementation


lib/browser/mqtt_internal/client.ts
```diff
@@ -0,0 +1,1181 @@
...
+        this.reevaluateService();
+    }
+
+    private transitionToStateConnected() {
+        this.changeState(ClientState.Connected);
+        this.linkToConnection();
+        let currentTime = this.getCurrentTime();
+        this.resetConnectionFailuresTimepoint = currentTime + (this.config.resetConnectionFailureCountMillis ?? DEFAULT_RESET_CONNECTION_FAILURE_COUNT_MILLIS);
+
+        let connackTimeout = this.pendingConnectionTimeout ?? this.creationTime;
```
sbSteveK: Trivial: It appears to be impossible for pendingConnectionTimeout to be undefined here. This can only be called from TransitionToStateConnecting -> transitionToState -> here and TransitionToStateConnecting sets pendingConnectionTimeout.

Outside of the undefined check, pendingConnectionTimeout is set using getCurrentTime() which determines elapsed time since creation (Date().getTime() - creationTime)  whereas creationTime is a straight Date().getTime() which will probably be an unexpectedly large number since other time values I'm seeing for scheduling appear to be use getCurrentTime(). Could be wrong about that, still processing.

lib/browser/mqtt_internal/client.ts
```diff
@@ -0,0 +1,1181 @@
...
+/**
+ * Type signature for a function that handles ConnectingEvent events
+ */
+export type ConnectingEventListener = (eventData: ConnectingEvent) => void;
+
+/**
+ * Emitted when the client successfully establishes an MQTT connection to the remote endpoint
+ */
+export interface ConnectionSuccessEvent {
+    connack: mqtt5_packet.ConnackPacket,
```
sbSteveK, line 42: No NegotiatedSettings?

lib/browser/mqtt_internal/client.ts
bretambrose: Could change it to zero rather than this.creationTime

lib/browser/mqtt_internal/client.ts
bretambrose: Not needed technically.  Even if we did include it, I'd make it a clone of the locally-computed copy so that protocol state wasn't externally modifiable.  Nothing stops us from adding it later if a reason comes up.

lib/browser/mqtt_internal/resubscribe.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    constructor() {
+
+    }
+
+    reset() {
+        this.subscriptions.clear();
+    }
+
+    onSubscribeRequest(topicFilter: string, qos: mqtt5_packet.QoS) {
```
sbSteveK, line 22: Edge case:
User is offline with OfflineQueuePolicy set to never. Subscribe stores this before checking the policy. It's discarded. Once connection is established we send out a subscribe based on `buildResubscribePacketList` that shouldn't be sent due to OfflineQueuePolicy.

We submit a subscription but it's rejected (maybe permissions) but the sub is set before we process the suback. On a reconnect we attempt to resub despite it having failed the first time so we aren't really "resubscribing". This initial failure that is followed later on in a discon/recon may be a success due to some kind of permission or broker change. This one is less an issue but still probably possible.

lib/browser/mqtt_internal/resubscribe.ts
```diff
@@ -0,0 +1,56 @@
...
+        this.subscriptions.set(topicFilter, qos);
+    }
+
+    onUnsubscribeRequest(topicFilter: string) {
```
sbSteveK, line 26: We could probably use this to remove a subscription that failed if we get a negative suback for the second issue listed above.

lib/browser/mqtt_internal/client.ts
```diff
@@ -0,0 +1,1181 @@
...
+        }
+
+        this.pendingConnectionTimeout = undefined;
+        if (event.packet.reasonCode == mqtt5_packet.ConnectReasonCode.Success) {
+            this.emit(Client.CONNECTION_SUCCESS, {
+                connack: event.packet
+            });
+        }
+
+        if (this.config.resubscribeMode == ResubscribeModeType.EnabledAlways ||
```
sbSteveK: We should move this into the ConnectReasonCode.Success above. It doesn't make sense to run it on a failed Connack.

sbSteveK (commented):
Some minor comments on the resubscribe stuff. Will move onto bulk of client next.

lib/browser/mqtt_internal/client.ts
```diff
@@ -0,0 +1,1181 @@
...
+
+    result += `${prefix}}`
+
+    return result;
+}
+
+function buildConnectionFailureEventLogString(event: ConnectionFailureEvent, prefix: string) : string {
+    let result = `${prefix}ConnectionFailureEvent: {\n`;
+
+    result += `${prefix}  error: "${event.error.toString()}\n"`;
```
sbSteveK: Utterly Trivial: The newline should be after the quotation is closed otherwise it'll print the close quote on the next line after the error is printed.

lib/browser/mqtt_internal/client.ts
```diff
@@ -0,0 +1,1181 @@
...
+            logDebug(CLIENT_LOG_SUBJECT, "clearing scheduled service task");
+            clearTimeout(this.serviceTask);
+            this.serviceTask = undefined;
+            this.nextServiceTimepoint = undefined;
+        }
+
+        if (serviceTime != undefined) {
+            let futureMillis = serviceTime - currentTime;
+            flogDebug(CLIENT_LOG_SUBJECT, () => { return `scheduling next service for ${futureMillis} millis from now`; });
+            setTimeout(this.service.bind(this), Math.max(0, futureMillis));
```
sbSteveK: I think this should probably be `this.serviceTask = setTimeout(...)`

Looking at https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout it returns a "timeout ID" which can be passed to `clearTimeout()` which is used above where we pass `this.serviceTask` to clear it. From what I can tell, we never assign anything but undefined to `this.serviceTask` and I think this is the most likely place where we are setting it to be checked and used.

Currently since `this.serviceTask` is ALWAYS undefined, the above `if (this.servicetask != undefined)` is always a no-op meaning that we never clear a timeout so every call to `reevaluateService` that reaches `setTimeout` schedules a new timer without cancelling the previous one. Tests are passing so maybe this isn't an issue or it's handled by `if (serviceTime != undefined && this.nextServiceTimepoint != undefined && serviceTime >= this.nextServiceTimepoint)`. Haven't thought it through beyond what this means but this is probably worth changing or removing as a no-op if it's unecessary.

lib/browser/mqtt_internal/client.ts
```diff
@@ -0,0 +1,1181 @@
...
+     * Always resubscribe to topics on reconnect, regardless of session resumption
+     */
+    EnabledAlways,
+
+    /**
+     * Only resubscribe to topics on reconnect if a session could not be rejoined
+     */
+    EnabledOnSessionResumptionFail,
+
+    Default = 0,
```
xiazhvera, line 120: The default mode seems not be used at all.

lib/browser/mqtt_internal/client.ts
```diff
@@ -0,0 +1,1181 @@
...
+    result = log.appendOptionalNumericPropertyLine(result, prefix, "MaxReconnectDelayMs", config.maxReconnectDelayMs);
+    result = log.appendOptionalNumericPropertyLine(result, prefix, "ResetConnectionFailureCountMillis", config.resetConnectionFailureCountMillis);
+    result = log.appendOptionalEnumPropertyLine(result, prefix, "ResubscribeMode", (val) => ResubscribeModeType[val], config.resubscribeMode);
+
+    result += `${prefix}}\n`;
+
+    return result;
+}
+
+function validateUnsignedInteger(value: number, fieldName: string) {
```
xiazhvera: Trivial: it feels cleaner having the validation help functions all managed at the same place in validate module. I'm talking about the help functions like validateUnsignedInteger / validateOptionalPositiveU32, not including client specific validation function like `validateConnectOptions`, as validate module seems to focus on the protocol related validation.

lib/browser/mqtt_internal/client.ts
bretambrose: was a mistake

lib/browser/mqtt_internal/resubscribe.ts
bretambrose: Disagree that this is an issue.  I think of this as capturing "intent".  If the user wants to subscribe to X we will try to subscribe to X.  If it fails, it fails, but by triggering off of subscribe/unsubscribe we stay true to original intent without having to track acks or additional bookkeeping that I don't think buys us anything useful.

lib/browser/mqtt_internal/client.ts
bretambrose: Sure, can move

lib/browser/mqtt_internal/client.ts
bretambrose: While it's internal, the basic intention of the client interface is to be as if it were public.

sbSteveK (approved):

xiazhvera:

PR #705, opened 2026-03-06
Js Mqtt - Pr #8 MQTT Public Client 311 Integration

bretambrose:
Migrates the public 311 client from mqtt-js to the new internal client


lib/browser/mqtt.ts
```diff
@@ -673,91 +731,39 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
-        /**
-         * This shouldn't ever occur, but in THEORY it could be possible to have on_disconnected called with the intent
-         * to disconnect without on_close called first. This would properly emit 'closed' should that unlikely event occur.
-         */
-        if (this.currentState == MqttBrowserClientState.Connected && this.desiredState == MqttBrowserClientState.Stopped) {
-            let closedCallbackData = {} as OnConnectionClosedResult;
-            this.emit("closed", closedCallbackData);
+        /* Did we intend to disconnect? If so, then emit the event */
+        if (this.desiredState == MqttBrowserClientState.Stopped) {
+            this.emit("closed");
```
sbSteveK: It appears as though if a user calls disconnect() we emit closed here and then a second time below in the on_stopped(). I don't think it's a bug since "closed" is an  intentional user driven event but we should still probably only emit once.

lib/browser/mqtt.ts
```diff
@@ -673,91 +731,39 @@ export class MqttClientConnection extends BufferedEventEmitter {
...
+    private on_connection_failure = (event: internal_mqtt_client.ConnectionFailureEvent) => {
+        let failureCallbackData = { error: event.error } as OnConnectionFailedResult;
+        this.emit('connection_failure', failureCallbackData);
     }
 
-    private on_disconnected = () => {
-        this.emit('disconnect');
+    private on_disconnection = (event : internal_mqtt_client.DisconnectionEvent)=> {
+        this.currentState = MqttBrowserClientState.Stopped;
+        this.emit('interrupt', -1);
```
sbSteveK, line 668: The static INTERRUPT defines this as emitted when connection is dropped unexpectedly. I think we can get here on a manual disconnect initiated by the user. Should we change this to not always report a -1? Or not emit if it's a user initiated disconnection?

bretambrose: The previous always emitted -1 for these events.  Why I have no idea.

As for conditionality, afaik, interrupt was always emitted for mqtt-js close events which are always emitted for network stream closes, user-initiated disconnects included.

lib/browser/mqtt.ts
bretambrose: Looks wrong.  Will poke around.

lib/browser/mqtt.ts
```diff
@@ -269,28 +273,28 @@ class TopicTrie extends Trie<OnMessageCallback | undefined> {
...
     private subscriptions = new TopicTrie();
     private connection_count = 0;
 
-    // track number of times in a row that reconnect has been attempted
-    // use exponential backoff between subsequent failed attempts
-    private reconnect_count = 0;
-    private reconnect_min_sec = DEFAULT_RECONNECT_MIN_SEC;
-    private reconnect_max_sec = DEFAULT_RECONNECT_MAX_SEC;
+    private connectPromise?: promise.LiftedPromise<boolean> = undefined;
+    private isSessionPresent: boolean = false;
```
sbSteveK, line 220: isSessionPresent is never changed from false. It should probably be set to true in `on_online`. Or removed if it's completely unecessary? I'll have to look and see where and how it's used and if it never being true is a problem.

lib/browser/mqtt.ts
```diff
@@ -12,31 +12,35 @@
  * @mergeTarget
  */
 
-import * as mqtt from "mqtt";
-import * as WebsocketUtils from "./ws";
+import * as internal_mqtt_client from "./mqtt_internal/client";
+import * as mqtt5 from "../common/mqtt5";
+import * as ws from "./ws"
+import * as WebsocketUtils from "./ws"
```
sbSteveK: Trivial: just above is `import * as ws from "./ws"` and I think that includes the WebsocketUtils that's being imported on this line.

sbSteveK (approved):

xiazhvera:

PR #706, opened 2026-03-11
Js Mqtt - Pr #9 MQTT5 Integration

bretambrose:
Integrate the internal MQTT client with our MQTT5 public client.


lib/browser/mqtt5.spec.ts
```diff
@@ -329,17 +330,40 @@ function createOperationFailureClient() : mqtt5.IMqtt5Client {
...
+/* Variant of the shared version that doesn't expect the stop call to throw */
+async function testDisconnectValidationFailure(client : mqtt5.Mqtt5Client, sessionExpiry: number) {
+    let connectionSuccess = once(client, mqtt5.Mqtt5Client.CONNECTION_SUCCESS);
+    let disconnection = once(client, mqtt5.Mqtt5Client.DISCONNECTION);
+
+    client.start();
+
+    await connectionSuccess;
+
+    client.stop({
```
xiazhvera: Does this means we no longer throw errors for stop operation. Would this be an API change?

lib/browser/mqtt5.ts
```diff
@@ -772,108 +601,100 @@ export class Mqtt5Client extends BufferedEventEmitter implements mqtt5.IMqtt5Cli
...
+    if (config == null || config == undefined) {
+        throw new CrtError("create_negotiated_settings: config not defined");
+    }
+    if (connack == null || connack == undefined) {
+        throw new CrtError("create_negotiated_settings: connack not defined");
+    }
+
+    return {
+        maximumQos: Math.min(connack.maximumQos ?? mqtt5_packet.QoS.ExactlyOnce, mqtt5_packet.QoS.AtLeastOnce),
+        sessionExpiryInterval: connack.sessionExpiryInterval ?? config.connectProperties?.sessionExpiryIntervalSeconds ?? 0,
```
sbSteveK, line 642: Trivial: "sessionExpiryInterval" missing time unit (Seconds)

bretambrose: These are all existing types, unsure what the issue is.

lib/browser/mqtt5.spec.ts
bretambrose: We do not throw, correct.  What problem would this cause?  For me, the motivation is to match intent.  If I say "stop" but misconfigure the disconnect packet, I think the client should still stop rather than fail-to-stop.

lib/browser/mqtt5.ts
sbSteveK: No real issue, just calling out that we don't have a time unit associated in this type. Maybe we can't change it because it'd be a breaking change

lib/browser/mqtt5.spec.ts
xiazhvera: Though the change make sense, I was worried this would be a breaking change for users. 
Also as you mentioned here, it is actually a behavior change on stop operation. I believe all other SDKs including node would validate the disconnect packet and fail-to-stop on a misconfig disconnect packet.

bretambrose: I've kept the internal client as is, but added a validation check for the mqtt5 client before submitting internally.

sbSteveK (approved):

xiazhvera:

PR #708, opened 2026-03-23
JS Mqtt - PR #10 Manual acknowledgements

bretambrose:
WIP

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #710, opened 2026-03-26
Manual Publish Acknowledgement

sbSteveK:
Bind manual publish acknowledgement feature for MQTT5 client

emitWithCallback() From ( https://github.com/awslabs/aws-crt-nodejs/pull/708 )

Manual acknowledgement tests.

### DIFF
--- a/lib/common/mqtt_shared.ts
+++ b/lib/common/mqtt_shared.ts
@@ -147,7 +147,7 @@ export class PublishAcknowledgementHandleWrapper {
         let handle = this.ackHandle;
         this.ackHandle = null;
 
-        return handle;``
+        return handle;
     }
 }
 
--- a/lib/common/mqtt_shared.ts
+++ b/lib/common/mqtt_shared.ts
@@ -122,3 +122,54 @@ export function isValidTopic(topic: any) : boolean {
 
     return isValidTopicInternal(topicAsString, false);
 }
+
+export type PublishAcknowledgementFunctor = () => void;
+
+/**
+ * Wrapper class containing a one-use singleton handle that can be used to trigger sending the acknowledgement (Puback in
+ * QoS 1, Pubrec in QoS 2) packet for an incoming publish.
+ */
+export class PublishAcknowledgementHandleWrapper {
+
+    private ackHandle : PublishAcknowledgementHandle | null;
+
+    constructor(acknowledgementFunction : PublishAcknowledgementFunctor) {
+        this.ackHandle = new PublishAcknowledgementHandle(acknowledgementFunction);
+    }
+
+    /**
+     * Attempt to take the acknowledgement handle held by the wrapper.  This will only succeed for the first caller;
+     * after the initial call, null will be returned.  By taking the handle, the caller assumes responsibility
+     * for sending the acknowledgement packet associated with the incoming publish packet.  Failing to trigger the
+     * acknowledgement will cause the broker to potentially re-send the publish.
+     */
+    acquireHandle() : PublishAcknowledgementHandle | null {
+        let handle = this.ackHandle;
+        this.ackHandle = null;
+
+        return handle;``
+    }
+}
+
+/**
+ * Object that allows the holder to trigger the acknowledgement for an associated publish packet.
+ */
+export class PublishAcknowledgementHandle {
+
+    private acknowledgementFunction? : PublishAcknowledgementFunctor;
+
+    constructor(acknowledgementFunction : PublishAcknowledgementFunctor) {
+        this.acknowledgementFunction = acknowledgementFunction;
+    }
+
+    /**
+     * trigger the acknowledgement for an associated Publish packet
+     */
+    invokeAcknowledgement() : void {
+        let acknowledgementFunction = this.acknowledgementFunction;
+        this.acknowledgementFunction = undefined;
+        if (acknowledgementFunction) {
+            acknowledgementFunction();
+        }
+    }
+}
--- a/lib/common/mqtt_shared.ts
+++ b/lib/common/mqtt_shared.ts
@@ -7,6 +7,8 @@
  * @packageDocumentation
  */
 
+import * as event from "./event";
+
 
 /**
  * Converts payload to Buffer or string regardless of the supplied type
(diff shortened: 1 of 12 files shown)


lib/native/mqtt5.spec.ts
```diff
@@ -12,7 +12,7 @@ import {v4 as uuid} from "uuid";
 import * as io from "./io";
 import {once} from "events";
 
-jest.setTimeout(30000);
+jest.setTimeout(40000);
```
bretambrose: Can probably restore this

lib/common/mqtt_shared.ts
```diff
@@ -121,4 +121,55 @@ export function isValidTopic(topic: any) : boolean {
...
+ */
+export class PublishAcknowledgementHandleWrapper {
+
+    private ackHandle : PublishAcknowledgementHandle | null;
+
+    constructor(acknowledgementFunction : PublishAcknowledgementFunctor) {
+        this.ackHandle = new PublishAcknowledgementHandle(acknowledgementFunction);
+    }
+
+    /**
```
bretambrose, line 142: This isn't the latest version.

lib/native/mqtt5.ts
```diff
@@ -662,13 +662,38 @@ export class Mqtt5Client extends NativeResourceMixin(BufferedEventEmitter) imple
...
+                crt_native.mqtt5_client_invoke_publish_acknowledgement(client.native_handle(), controlId);
+            };
+            acknowledgementControl = new mqtt_shared.PublishAcknowledgementHandleWrapper(functor);
+        }
+
         let messageReceivedEvent: mqtt5.MessageReceivedEvent = {
             message: message
         };
 
+        if (acknowledgementControl) {
```
bretambrose, line 681: update to use queueAcknowledgeableEvent

source/mqtt5_client.c
```diff
@@ -312,6 +314,17 @@ static void s_on_publish_received(const struct aws_mqtt5_packet_publish_view *pu
         return;
     }
 
+    /*
+     * Eagerly acquire manual publish acknowledgement control for QoS 1 publishes.  This MUST be called synchronously
+     * before this function returns.  A return value of 0 means the acquire
+     * failed or the publish is not QoS 1.
+     */
+    uint64_t puback_control_id = 0;
+    if (publish_packet->qos == AWS_MQTT5_QOS_AT_LEAST_ONCE) {
```
bretambrose: would prefer `!= AT_MOST_ONCE`

source/mqtt5_client.c
```diff
@@ -1219,6 +1232,22 @@ static void s_napi_on_message_received(napi_env env, napi_value function, void *
             goto done;
         }
 
+        /*
+         * Pass the puback_control_id as a napi_external wrapping a heap-allocated uint64_t*
+         * (due to no napi_create_bigint_uint64 available). A value of 0 means QoS 0 or acquire failed
```
bretambrose: Our minimum version is now greater than the version that added this.  Should we not use externals here?

lib/native/mqtt5.spec.ts
sbSteveK: Will revert this to 30000 and manually change them in the tests that run longer below to 45 seconds.

source/mqtt5_client.c
sbSteveK: Works for me. Changed.

source/mqtt5_client.c
sbSteveK: I think this was failing locally because I may have setup my dev env to whatever the min version was before. We can probably use an external but this also feels safer in terms of keeping the value fully opaque to the user. If you feel strongly, I can change it but I think its current implementation better fits what we're trying to do in keeping it out of user view. I'll update the comment about it not being available.

bretambrose: Let's add a backlog issue to bump our min napi version to in-sync with min node version.

lib/common/mqtt_shared.ts
sbSteveK: Updating to what's current in native client pr

lib/native/mqtt5.ts
sbSteveK: Changed to use the new updated shared queueAcknowledgeableEvent

source/mqtt5_client.c
```diff
@@ -3615,6 +3643,65 @@ napi_value aws_napi_mqtt5_client_get_queue_statistics(napi_env env, napi_callbac
...
+    });
+
+    if (control_id_ptr == NULL) {
+        napi_throw_error(
+            env, NULL, "aws_napi_mqtt5_client_invoke_publish_acknowledgement - control id pointer was null");
+        return NULL;
+    }
+
+    uint64_t control_id = *control_id_ptr;
+    aws_mem_release(aws_napi_get_allocator(), control_id_ptr);
```
bretambrose: Concerned about object/reference leaks here.  This isn't the right place to be destroying this.

source/mqtt5_client.c
```diff
@@ -1219,6 +1232,21 @@ static void s_napi_on_message_received(napi_env env, napi_value function, void *
...
         }
 
+        /*
+         * Pass the puback_control_id as a napi_external wrapping a heap-allocated uint64_t*.
+         * A value of 0 means QoS 0 or acquire failed and we pass undefined.
+         */
+        if (on_message_received_ud->puback_control_id != 0) {
+            uint64_t *control_id_ptr = aws_mem_calloc(aws_napi_get_allocator(), 1, sizeof(uint64_t));
+            *control_id_ptr = on_message_received_ud->puback_control_id;
+            AWS_NAPI_CALL(env, napi_create_external(env, control_id_ptr, NULL, NULL, &params[2]), {
```
bretambrose: Why is this created without a destructor?

sbSteveK: added destructor to clean up the uint64_t

source/mqtt5_client.c
sbSteveK: Added a finalizer that should clean up the memory allocated to control_id_ptr. Removed this explicit release as it's redundant.

bretambrose (approved):

PR #717, opened 2026-05-22
Support s2n-tls on macOS

sfod:
Add support for s2n-tls as a TLS backend on macOS.
Apple Secure Transport remains the default for FIPS compliance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,10 @@ if (BUILD_DEPS)
     set(BUILD_TESTING OFF)
 
     add_subdirectory(crt/aws-c-common)
-    if (UNIX AND NOT APPLE)
+    # On Linux and BSD, we use S2N for TLS and AWS-LC crypto.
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -49,6 +53,7 @@ if (BUILD_DEPS)
             CMAKE_ARGUMENTS
                 -DUNSAFE_TREAT_WARNINGS_AS_ERRORS=OFF
                 -DBUILD_TESTING=OFF
+                -DCMAKE_NO_SYSTEM_FROM_IMPORTED=ON
```
xiazhvera: It looks like we always manually enable `CMAKE_NO_SYSTEM_FROM_IMPORTED` here, but according to the comment in CI, we only need it for x86. I guess it is fine to set `CMAKE_NO_SYSTEM_FROM_IMPORTED` for s2n, but it is a little confusing as we have different setup across the libraries.

xiazhvera (approved):

PR #719, opened 2026-06-03
Remove excess cmake flag from macOS-x64 CI

sfod:
Remove an excess cmake option from macos-x64 CI jobs.
This option is set in CMakeLists.txt, which is more robust since it doesn't require passing it explicitly to build process.


azkrishpy:
we should probably document why the flag is not needed anymore as well since there seems to have been a detailed comment on why this flag was needed on intel macs

sfod:
The flag is set in CMakeLists.txt, with similar comments. I updated the description in this PR.

rakshil14-2 (approved):
LGTM

PR #724, opened 2026-06-19
Js Mqtt - PR #12 metrics improvement for native layer

rakshil14-2:
*Description of changes:*
  - Added an `encoded feature list` system for IoT SDK metrics that reports client configuration choices (retry jitter, session behavior, offline queue behavior, topic aliasing, TLS settings, proxy type, socket implementation) as compact ID/Value pairs in the MQTT CONNECT packet.
  - Added metrics support for `native Node.js` MQTT3 and MQTT5 clients via createMetricsMqtt5 / createMetricsMqtt3.
  - Moved `ClientOperationQueueBehavior` from `lib/native/mqtt5.ts` to `lib/common/mqtt5.ts` to avoid circular dependency.
  - Used `type` to import client type in `aws_iot_metrics`
  - Refactored the C native layer to use a shared `aws_napi_metrics_parse` helper (in module.c) that parses the full metrics object (library name + metadata key/value pairs) for both MQTT3 and MQTT5 clients.
  - Added `Certificate source` enum to gather information about certificate source
  - Added unit tests for all feature value mappers, encoding functions, and merge logic.
  - Added Factory pattern for IoT SDK layer registration : `_setSDkMetricsFactory()`


lib/native/aws_iot_metrics.ts
```diff
@@ -0,0 +1,502 @@
...
+}
+
+/**
+ * Map protocol version to its single-character metrics value.
+ *
+ * Mapping: MQTT311->3, MQTT5->5.
+ * Always emitted because every connection has a known protocol version.
+ * @internal
+ */
+export function protocol_version_metrics_value(protocol: "MQTT311" | "MQTT5"): string {
```
xiazhvera: Let's make a constant/enum instead of hardcoded string for mqtt311 and mqtt5.

lib/native/mqtt5.ts
```diff
@@ -206,6 +178,15 @@ export interface Mqtt5ClientConfig extends mqtt_shared.Mqtt5ClientConfigBase {
...
     extendedValidationAndFlowControlOptions? : ClientExtendedValidationAndFlowControl;
+
+    /**
+     * Optional metrics configuration for IoT SDK metrics reporting.
+     * If provided, the CRT will merge with CRT-generated metrics.
+     * If undefined, default metrics will be created.
+     *
+     * @group Node-only
+     */
+    metrics?: mqtt_shared.AwsIoTDeviceSDKMetrics;
```
xiazhvera: As disable metrics was defined in Mqtt5ClientConfigBase, should the metrics move to Mqtt5ClientConfigBase as well?  Similar to MqttConnectionConfigBase.

source/module.c
```diff
@@ -1618,3 +1620,225 @@ static bool s_create_and_register_function(
...
+ *   { libraryName: "IoTDeviceSDK/JS", metadata: [["CRTVersion","1.0.0"], ["IoTSDKFeature","F/5,G/A"], ...] }
+ *
+ * Caller must call aws_napi_metrics_clean_up(storage) after the metrics are no longer needed.
+ * Shared by both MQTT5 (mqtt5_client.c) and MQTT3 (mqtt_client_connection.c).
+ *
+ * Returns AWS_OP_SUCCESS if metrics were successfully parsed (out_metrics is usable),
+ * or AWS_OP_ERR if the metrics object was null/invalid (out_metrics should not be used).
+ */
+
+int aws_napi_metrics_parse(
```
xiazhvera: I dont think module is a good place for metrics parsing. module file is more module level functions and common utils. metrics is specific for IoT metrics. I would prefer it in mqtt related files or it's own file.

source/module.c
```diff
@@ -1618,3 +1620,225 @@ static bool s_create_and_register_function(
...
+
+    AWS_LOGF_DEBUG(AWS_LS_NODEJS_CRT_GENERAL, "aws_napi_metrics_parse: Creating metrics from JS object");
+
+    /*
+     * Extract the library name (e.g. "IoTDeviceSDK/JS").
+     * Copies the JS string into library_name_buf and sets metrics.library_name
+     * as a cursor pointing into that buffer.
+     */
+    napi_value node_library_name = NULL;
+    if (napi_get_named_property(env, node_metrics, "libraryName", &node_library_name) != napi_ok ||
```
xiazhvera: Let's use constant var instead of hard coded string here. refer to:  source/mqtt_request_response.c

lib/native/aws_iot.ts
```diff
@@ -467,6 +469,31 @@ export class AwsIotMqttConnectionConfigBuilder {
...
     }
 
+    /**
+     * Provides SDK-level metrics to be sent in the MQTT CONNECT packet's
+     * username field. The CRT will merge these with its own auto-detected
+     * transport-level feature metrics and embed the combined result.
+     *
+     * @param metrics SDK-level metrics containing library name and metadata
+     */
+    with_metrics(metrics: mqtt_shared.AwsIoTDeviceSDKMetrics) {
```
xiazhvera: Builder/IoT API should not expose any metrics related features if not pre-exists.

lib/native/aws_iot_mqtt5.ts
```diff
@@ -479,6 +480,32 @@ export class AwsIotMqtt5ClientConfigBuilder {
...
 
+    /**
+     * Provides SDK-level metrics to be sent in the MQTT CONNECT packet's
+     * username field. The CRT will merge these with its own auto-detected
+     * transport-level feature metrics (TLS settings, certificate source,
+     * proxy, retry behavior, etc.) and embed the combined result.
+     *
+     * @param metrics SDK-level metrics containing library name and metadata
+     */
+    withMetrics(metrics: mqtt_shared.AwsIoTDeviceSDKMetrics) : AwsIotMqtt5ClientConfigBuilder {
```
xiazhvera: Similar to the above comment. We dont expose metrics in IoT/Builders.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * Returns AWS_OP_SUCCESS if metrics were successfully parsed (out_metrics is usable),
+ * or AWS_OP_ERR if the metrics object was null/invalid (out_metrics should not be used).
+ */
+
+int aws_napi_metrics_parse(
+    napi_env env,
+    napi_value node_metrics,
+    struct aws_mqtt_iot_metrics *out_metrics,
+    struct aws_napi_metrics_storage *storage) {
```
xiazhvera: You want to explicitly call out that storage should be cleanup before pass in the function, otherwise ZERO_STRUCTURE will cause a memory leak. Also out_storage for name conversion to indicate that the function will modify the storage param.

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,233 @@
...
+
+    /*
+     * Extract the library name (e.g. "IoTDeviceSDK/JS").
+     * Copies the JS string into library_name_buf and sets metrics.library_name
+     * as a cursor pointing into that buffer.
+     */
+    napi_value node_library_name = NULL;
+    if (napi_get_named_property(env, node_metrics, AWS_NAPI_KEY_LIBRARY_NAME, &node_library_name) != napi_ok ||
+        aws_byte_buf_init_from_napi(&storage->library_name_buf, env, node_library_name) != AWS_OP_SUCCESS) {
+        AWS_LOGF_DEBUG(AWS_LS_NODEJS_CRT_GENERAL, "aws_napi_metrics_parse - failed to read libraryName");
```
xiazhvera: Should it be error log here? Also, do we want to raise error like `aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);`?

source/mqtt_iot_metrics.c
```diff
@@ -0,0 +1,233 @@
...
+    struct aws_mqtt_iot_metrics *out_metrics,
+    struct aws_napi_metrics_storage *storage) {
+
+    /* Zero-initialize so all fields are safe for cleanup on any error path */
+    AWS_ZERO_STRUCT(*out_metrics);
+    AWS_ZERO_STRUCT(*storage);
+
+    if (aws_napi_is_null_or_undefined(env, node_metrics)) {
+        AWS_LOGF_DEBUG(AWS_LS_NODEJS_CRT_GENERAL, "aws_napi_metrics_parse - metrics object is null/undefined");
+        return AWS_OP_ERR;
```
xiazhvera: It's a little wired that it returns error but actually it is a valid case. I think it makes more sense just return success with out_metrics set to nullptr.

xiazhvera (changes requested):
I revisit the metrics binding and left some comments.

lib/native/aws_iot.ts
```diff
@@ -467,6 +468,19 @@ export class AwsIotMqttConnectionConfigBuilder {
...
 
+    /**
+     * Sets whether to disable IoT SDK metrics.
+     *
+     * When disabled, no metrics (SDK or CRT) are included in the CONNECT packet.
+     * Defaults to false (metrics enabled).
+     *
+     * @param disable true to disable metrics, false to enable (default)
+     */
+    with_disable_metrics(disable: boolean) {
```
xiazhvera: Also remove this, for both here and mqtt5.

lib/common/aws_iot_shared.ts
```diff
@@ -242,9 +240,6 @@ export function buildMqtt5FinalUsername(customAuthConfig?: MqttConnectCustomAuth
         }
     }
 
-    paramList.push(["SDK", SDK_NAME]);
```
xiazhvera, line 245: Note: Removing this might remove the username params from browser implementation as well. We would need make sure this get covered with future browser metrics support.

lib/native/aws_iot.ts
```diff
@@ -485,7 +486,7 @@ export class AwsIotMqttConnectionConfigBuilder {
                 }
             }
         }
-
+
```
xiazhvera: Clean up.

lib/native/aws_iot_mqtt5.ts
```diff
@@ -478,7 +479,7 @@ export class AwsIotMqtt5ClientConfigBuilder {
         this.config.topicAliasingOptions = topicAliasingOptions;
         return this;
     }
-
+
```
xiazhvera: Need clean up.

xiazhvera (approved):
Clean up& approved. The PR should ship with browser implementation.

Issue #607, opened 2025-03-17
Replace axios with built-in alternatives


kuhe:
### Describe the feature

Can use of axios be replaced by native node:http(s), fetch, or XMLHttpRequest?

### Use Case

Because this package (aws-crt) is a dependency of the AWS SDK for JavaScript (v3), SDK users are affected by recurrent vulnerabilities discovered in the axios package. 

https://security.snyk.io/package/npm/axios

Secondarily, it would also reduce the package size to remove this dependency.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

trivikr:
Revisiting this issues from another ticket to JS SDK team (internal JS-5564)

Axios is used in `lib/browser/http.ts`, and it can be easily be replaced with [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)
https://github.com/awslabs/aws-crt-nodejs/blob/7b3fc6ccf709cffe733c067dd4bdc3defc760d6b/lib/browser/http.ts#L30

JS SDK uses fetch in [fetch-http-handler](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/fetch-http-handler/src/fetch-http-handler.ts).

There are codemods available to perform migration.
For example, [Axios → Fetch with Codemod 2.0](https://codemod.com/blog/axios-fetch-with-codemod-2-0).

trivikr:
On 3/31, axios@1.14.1 and axios@0.30.4 were compromised with a supply chain attack
Details in https://socket.dev/blog/axios-npm-package-compromised

It would have impacted aws-crt-nodejs users who don't use lockfile, and/or updated their lockfile
https://github.com/awslabs/aws-crt-nodejs/blob/a201f639f408f6c85a3c378c7c99711fe82fef8a/package.json#L57

npm quicky responded and removed the malicious dependency.
But to ensure it doesn't happen again, libraries should consider replacing axios with built-in alternatives.

bretambrose:
It is not clear how to emulate the proxy support that axios provides.  Absent a solution there, we would have to take the risk that no one is using an http proxy and the browser implementation in node.  I doubt anyone is doing that, but I have no way of actually knowing, hence the reluctance to move further with this.

TrevorBurnham:
[...]

PR #730, opened 2026-07-18
Replace axios with fetch, restore proxy support via undici

TrevorBurnham:
Replaces axios with `fetch` (browser HTTP impl) and undici (build scripts), and restores the HTTP proxy support that a naive fetch migration would drop.

Addresses the concern raised in #607: rather than betting nobody proxies through the browser impl under Node, proxy options are honored by routing `fetch` through undici's `ProxyAgent`. This is an alternative to #725.

## How proxy support works
- When running under Node.js with `proxy_options` set, requests go through undici's `ProxyAgent` via fetch's `dispatcher` option (Basic auth supported). IPv6 proxy hosts are bracketed; proxy construction failures surface through the connection's `error` event rather than a constructor throw, and `HttpClientConnectionManager` recovers its slot when that happens.
- Proxying under Node fails closed: if `proxy_options` are set but undici can't be loaded, the connection errors instead of silently connecting directly (which would bypass egress controls).
- Proxy options that only the native impl can honor (`tls_opts`, forwarding `connection_type`) emit a warning instead of being silently ignored.
- `undici` is lazily `require`d through an indirection (so bundlers don't statically resolve it) and stubbed out of browser bundles through the `package.json` `browser` field; real-browser builds never pull it in (verified with a webpack bundle: undici is reported as `ignored`).
- In a real browser, per-request proxying isn't possible; options have no effect and emit a warning.
- Build scripts (`build_step_http.js`) honor `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` via undici's `EnvHttpProxyAgent`, preserving the env-proxy behavior axios provided; these also fail closed if the configured proxy is unusable. (Axios additionally read `ALL_PROXY` and `npm_config_*` proxy vars; those are not carried over.)

## Behavior preserved from the axios implementation
- Request URLs are built by textual base+path join (axios `combineURLs` semantics), not `new URL(path, base)` — URL resolution would let a path starting with `//` override the connection's host and would normalize dot segments out of signed paths.
- Plain-object request bodies are JSON-serialized (with `content-type: application/json` defaulted), matching what the `BodyData` contract implies; `fetch` would otherwise coerce them to `"[object Object]"`. Bodies on GET/HEAD can't be expressed in fetch and now warn instead of silently dropping.

## Compatibility / semver
- Node floor rises to 18.17 (global `fetch`; undici v6 requires >= 18.17), enforced via a package.json `engines` field so users on older Node get an install-time signal instead of a runtime `ReferenceError`.
- Keeps the `overrides: { axios }` block so the remaining transitive dev dependency stays on a patched axios rather than reverting to a vulnerable range.
- `package-lock.json` kept in v1 format; `undici` v6 adds zero transitive deps.

## CI
- The Linux compat matrix keeps al2012/al2 on Node they can actually run (glibc < 2.28 can't load official Node 18 binaries): al2012 stays pinned to node-14 and al2 to node-16 via builder host overrides. These jobs only build/test the native addon.
- The puppeteer/browser-test host config moves from al2 to al2023, and `al2023-x64` is added to the CI matrix so `npm run test:browser:ci` still runs somewhere.

## Tests
- New `lib/browser/http_proxy.spec.ts`: stands up an in-process origin server + CONNECT-tunneling proxy and verifies requests transit the proxy, Basic `Proxy-Authorization` is sent correctly, and the no-proxy path bypasses it. Also covers the failure paths: an unusable proxy host surfaces via the `error` event, and a failed proxy connection doesn't permanently consume an `HttpClientConnectionManager` slot.
- The browser jest environment injects `fetch`/`TextDecoder` and the web-stream globals into the Jest 27 sandbox (undici won't load without `ReadableStream` et al. — without this, proxying silently degrades to direct connections in the test run).


xiazhvera:
Hi, as discussed in https://github.com/awslabs/aws-crt-nodejs/issues/607, once we make the formal announcement and complete the deprecation period, we will be able to bump the minimum supported version to Node 18 and removed Axios support. 
We will close the PR.

PR #732, opened 2026-07-22
FIx(ci): crash due to unhandled Promise 

rakshil14-2:
*Issue #, if available:*

*Description of changes:*
- An unhandled promise created a warning instead of failure due to `Node 14` default behavior that closed the socket silently


xiazhvera (changes requested):
It seems that we have unhandled promise here as well https://github.com/awslabs/aws-crt-nodejs/blob/main/samples/browser/pub_sub/index.ts#L93. Let's clean up other unhandled promises across the library.

xiazhvera (approved):

PR #733, opened 2026-07-27
Increase pacakge size

xiazhvera:
*Issue #, if available:*

*Description of changes:*
To prevent the size validation from blocking the CD pipeline, we updated the CD size validation. Instead of matching the current exact size, set the limit to a larger NPM package size. This acts validate against exceeding NPM's upload limit without being overly restrictive in CD, as CI already handles the size detection on each platform. 
NPM didn't have an hard limit for the package size, however, the Node.js CLI memory limits create an effective maximum threshold. (https://github.com/npm/npm/issues/12750 The thread is a little old, while I didn't find any update regards to the size limit.), so use 100 MB size limit as a safe threadhold.

The package size is increased from:
```
npm notice === Tarball Details ===
--
npm notice name:          aws-crt
npm notice version:       1.32.3
npm notice filename:      aws-crt-1.32.3.tgz
npm notice package size:  15.5 MB
npm notice unpacked size: 43.1 MB
```
to

```
npm notice === Tarball Details === 
npm notice name:          aws-crt                                 
npm notice version:       1.33.0                                  
npm notice filename:      aws-crt-1.33.0.tgz                      
npm notice package size:  15.7 MB                                 
npm notice unpacked size: 44.3 MB                               
```

### DIFF
--- a/continuous-delivery/pack.sh
+++ b/continuous-delivery/pack.sh
@@ -39,7 +39,7 @@ mkdir $UNZIP
 tar -xf aws-crt-$CURRENT_TAG.tgz -C $UNZIP
 PACK_FILE_SIZE_KB=$(du -sk $UNZIP | awk '{print $1}')
 echo "Current package size: ${PACK_FILE_SIZE_KB}"
-if expr $PACK_FILE_SIZE_KB \> "$((43000))" ; then
+if expr $PACK_FILE_SIZE_KB \> "$((45000))" ; then
     # the package size is too large, return -1
     echo "Package size is too large!"
     exit -1
--- a/continuous-delivery/pack.sh
+++ b/continuous-delivery/pack.sh
@@ -35,13 +35,20 @@ cp aws-crt-*.tgz ..
 # Check unzip npm package size
 cd ..
 UNZIP="unzip_pack"
+
+# NPM does not have an hard limit for the package size, however, the Node.js CLI memory limits 
+# create an effective maximum threshold. Reference to https://github.com/npm/npm/issues/12750. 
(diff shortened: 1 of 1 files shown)


continuous-delivery/pack.sh
```diff
@@ -39,7 +39,7 @@ mkdir $UNZIP
 tar -xf aws-crt-$CURRENT_TAG.tgz -C $UNZIP
 PACK_FILE_SIZE_KB=$(du -sk $UNZIP | awk '{print $1}')
 echo "Current package size: ${PACK_FILE_SIZE_KB}"
-if expr $PACK_FILE_SIZE_KB \> "$((43000))" ; then
+if expr $PACK_FILE_SIZE_KB \> "$((45000))" ; then
```
rakshil14-2: Can we have dynamic way to check ?
npm's registry already stores the unpacked size of every published version. So can we fetch the previous release's real size and gate on growth rather than an absolute magic number? Like allowing 5-10% increment in size from previous release

rakshil14-2 (approved):

continuous-delivery/pack.sh
sbSteveK: Let's change CD to the maximum size allowed by npm for the package size. That way when something fails to be published to npm, we are aware of it being expected to fail instead of silently failing and not knowing a version isn't in npm till someone tries to grab it.

Outside of the NPM limit, we don't need this check since we check for smaller file size increases in our regular CI. Removing a number we are increasing without knowing what it'll be prevents us from having a release in github not having a paired release on npm.

rakshil14-2 (approved):

PR #736, opened 2026-08-19
Warning during client creation

rakshil14-2:
*Description of changes:*
- Till January 2027, customer using aws-crt client with an end-of-life version of `Node.js` they will get a warning to upgrade to Node.js LTS version
- The warning is emitted on client creation.


lib/common/node_deprecation_warning.ts
```diff
@@ -0,0 +1,86 @@
+/*
```
bretambrose: Why would we put this file in common?

Feels like we could just drop it into `native/crt.ts`

lib/index.ts
```diff
@@ -38,6 +39,7 @@ export {
     mqtt,
     mqtt5,
     mqtt_request_response,
+    node_deprecation_warning,
```
bretambrose: Why export?

lib/common/node_deprecation_warning.ts
```diff
@@ -0,0 +1,86 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+/**
+ * Emits a one-time warning when aws-crt runs on a Node.js version that will be
+ * unsupported starting January 2027. Suppress it in any of these ways:
+ * @example
+ * require('aws-crt').node_deprecation_warning.suppress = true;
+ * @example
+ * globalThis.AWS_CRT_NODEJS_SUPPRESS_NODE_DEPRECATION_WARNING = true;
```
bretambrose: What is the value of having two different ways of suppressing this?

lib/common/node_deprecation_warning.ts
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+        if (!platform.is_nodejs() || typeof process.emitWarning !== 'function') {
+            return;
+        }
+        if (hasWarned || !isUnsupportedNodeVersion()) {
+            return;
+        }
+
+        hasWarned = true;
```
bretambrose: confused why we need both hasWarned and suppress

lib/common/node_deprecation_warning.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const node_deprecation_warning: NodeDeprecationWarningController = {
+    suppress: false,
+
+    emitWarning(): void {
+        if (node_deprecation_warning.suppress || isSuppressedExternally()) {
+            return;
+        }
+        if (!platform.is_nodejs() || typeof process.emitWarning !== 'function') {
```
bretambrose: Are we worried about this being the case?

lib/common/node_deprecation_warning.ts
bretambrose: I'm not sure I see the value in additional suppression mechanisms.  And the require shouldn't work because node_deprecation_warning isn't an exported symbol.

Let's stick with environment variables.

lib/common/node_deprecation_warning.ts
rakshil14-2: `hasWarned` flag: is to emit warning exactly once. If two clients are simultaneously created only once warning should be emited

lib/common/node_deprecation_warning.ts
rakshil14-2: yes we can move to it

lib/index.ts
rakshil14-2: It was only exported to expose the programmatic suppress

lib/common/node_deprecation_warning.ts
bretambrose: Why not just combine?

bretambrose (approved):

PR #737, opened 2026-08-30
Preserve literal types for event name constants

Peter-Schorn:
## Summary

This PR preserves the literal types of static event-name constants so they can be used with the library's existing typed `on()` overloads.

For example, this currently fails to compile:

```ts
mqttClient.on(
    mqtt5.Mqtt5Client.CONNECTION_SUCCESS,
    onSuccess
);
```

with:

```text
No overload matches this call.
  The last overload gave the following error.
    Argument of type 'string' is not assignable to parameter of type '"stopped"'.
mqtt5.d.ts(365, 5): The last overload is declared here.
```

`Mqtt5Client.CONNECTION_SUCCESS` has the value `"connectionSuccess"`, and `Mqtt5Client.on()` already accepts that exact literal:

```ts
on(
    event: 'connectionSuccess',
    listener: mqtt5.ConnectionSuccessEventListener
): this;
```

The issue is that the static constant is typed as `string`, so TypeScript cannot match it to the corresponding overload.

This PR narrows these constants to their actual literal values, for example:

```ts
static CONNECTION_SUCCESS : 'connectionSuccess' = 'connectionSuccess';
```

The same issue appears in several event-emitting classes, so the fix is applied consistently across the affected browser and native implementations.

Nearly all of the event-name values changed here are already exposed as exact string literals by the corresponding typed `on()` overloads. This change therefore aligns the static constants with the library's _existing_ public type API and allows consumers to use the provided constants with those overloads as intended. Narrowing the static constants therefore aligns two existing parts of the API rather than exposing new event-name values.

### DIFF
--- a/lib/browser/http.ts
+++ b/lib/browser/http.ts
@@ -273,21 +273,21 @@ export class HttpClientConnection extends BufferedEventEmitter {
      *
      * @event
      */
-    static CONNECT = 'connect';
+    static CONNECT : 'connect' = 'connect';
 
     /**
      * Emitted when an error occurs on the connection
      *
      * @event
      */
-    static ERROR = 'error';
+    static ERROR : 'error' = 'error';
 
     /**
      * Emitted when the connection has completed
      *
      * @event
      */
-    static CLOSE = 'close';
+    static CLOSE : 'close' = 'close';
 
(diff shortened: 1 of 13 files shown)


rakshil14-2 (commented):
LGTM

bretambrose (approved):

